Issue #72, opened 2014-08-17
Add post-merge, post-commit, and post-rebase hooks for indexing tags


lencioni:
The post-checkout index tags hook works well. But, I was recently reading [an article](http://tbaggery.com/2011/08/08/effortless-ctags-with-git.html) where the author set up git hooks to do this for post-merge, post-commit, and post-rebase (via a special case in post-rewrite).

I was thinking that this might be nice to have as well.

sds:
Implemented in 88cd039a72dcc85e04c88f825f799f01635ae1f1.

PR #92, opened 2015-01-14
Add support for post-commit hooks

jawshooah:
The `PostCommit` `Hook` and `HookContext` are pretty sparse, but I'm not sure what other information would be useful.

Refs #72.


lib/overcommit/git_repo.rb
```diff
@@ -20,8 +20,9 @@ def extract_modified_lines(file_path, options)
       lines = Set.new
 
       flags = '--cached' if options[:staged]
+      refs  = 'HEAD~ HEAD' if options[:last_commit]
```
sds: I don't see the need to limit the utility of this option to only allow retrieving the last commit. Let's change it to `options[:refs]` for now and pass it directly.

lib/overcommit/git_repo.rb
```diff
@@ -35,6 +36,15 @@ def extract_modified_lines(file_path, options)
       lines
     end
 
+    # Extract the set of modified lines from a given file.
+    #
+    # @param file_path [String]
+    # @param options [Hash]
+    # @return [Set] line numbers that have been modified in file
+    def extract_modified_lines_last_commit(file_path)
```
sds: There doesn't appear to be a compelling need to make this a separate helper. The canonical `extract_modified_lines` method will do just fine inline as it is only consumed by helpers anyway.

lib/overcommit/git_repo.rb
```diff
@@ -42,12 +52,20 @@ def extract_modified_lines(file_path, options)
     # @return [Array<String>] list of absolute file paths
     def modified_files(options)
       flags = '--cached' if options[:staged]
+      refs  = 'HEAD~ HEAD' if options[:last_commit]
```
sds: Ditto here about passing `refs` directly as an option.

lib/overcommit/git_repo.rb
```diff
@@ -42,12 +52,20 @@ def extract_modified_lines(file_path, options)
...
-      `git diff --name-only -z --diff-filter=ACM --ignore-submodules=all #{flags}`.
+      `git diff --name-only -z --diff-filter=ACM --ignore-submodules=all #{flags} #{refs}`.
         split("\0").
         map { |relative_file| File.expand_path(relative_file) }
     end
 
+    # Returns the names of all files that were modified in the last commit.
+    #
+    # @return [Array<String>] lsit of absolute file paths
+    def modified_files_last_commit
```
sds: No need to make this a separate helper.

lib/overcommit/hook_context/post_commit.rb
```diff
@@ -0,0 +1,20 @@
+module Overcommit::HookContext
+  # Contains helpers related to contextual information used by post-commit
+  # hooks.
+  class PostCommit < Base
+    # Get a list of files that were added, copied, or modified in the last
+    # commit. Renames and deletions are ignored, since there should be nothing
+    # to check.
+    def modified_files_last_commit
```
sds: I would prefer this to just be called `modified_files` to match that of the `PreCommit` hook context. The goal with these context classes was to provide a somewhat common API across all types of hooks where it made sense to do so. In this case, it makes sense that `modified_files` would only apply to the last commit when in a `post-commit` context.

lib/overcommit/hook_context/post_commit.rb
```diff
@@ -0,0 +1,20 @@
...
+      @modified_files_last_commit ||= Overcommit::GitRepo.modified_files_last_commit
+    end
+
+    # Returns the set of line numbers corresponding to the lines that were
+    # changed in a specified file.
+    def modified_lines_in_file_last_commit(file)
```
sds: Ditto on changing this to match the `PreCommit` hook context helper method, `modified_lines_in_file`.

sds:
Thanks for contributing a new type of hook, @jawshooah! I'm excited to include this in the project.

I've left some comments inline about some issues with consistency, but overall this looks good.

The specs are failing because you also need to manually add the `post-commit` hook to a couple of places: the `template-dir` in the repo, and in the failing specs.

I'd also like to see a `spec/overcommit/hook_context/post_commit_spec.rb` added. You can use the [PostCheckout](https://github.com/causes/overcommit/blob/master/spec/overcommit/hook_context/post_checkout_spec.rb) hook context spec as an example.

Once my comments and those issues are fixed, I'll be happy to merge this!

jawshooah:
@sds Thanks for the speedy response! I'll make the changes you suggested when I get the chance.

jawshooah:
@sds Sorry for the wait, the last month has been a busy one. I've made the changes you suggested, and am now working on adding specs.

jawshooah:
@sds Is this good to go as well? Should I squash my commits?

sds:
Merged with some adjustments in 15ce4c2 and c124647.

sds:
Also fixed the specs to work with older git versions in bae60c60b401c4fa9cbe068810596cdcb0e685dc.

PR #93, opened 2015-01-14
Add git-guilt post-commit hook

jawshooah:
Change-Id: I70d5c21b6507e6562221218aadd02a74a882c0dc

[git-guilt](https://www.npmjs.com/package/git-guilt) reports the change in blame since the last commit.

Blocked by #92

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -262,3 +262,14 @@ CommitMsg:
 
   TrailingPeriod:
     description: 'Checking for trailing periods in subject'
+
+PostCommit:
+  ALL:
+    requires_files: false
+    required: false
+    quiet: false
+
+  GitGuilt:
+    description: 'Calculating changes in blame since last commit'
+    required_executable: 'git-guilt'
+    install_command: 'npm install -g git-guilt'
--- a/lib/overcommit/hook/post_commit/git_guilt.rb
+++ b/lib/overcommit/hook/post_commit/git_guilt.rb
@@ -1,0 +1,9 @@
+module Overcommit::Hook::PostCommit
+  # Calculates the change in blame since the last revision.
+  class GitGuilt < Base
+    def run
+      system(executable, 'HEAD~', 'HEAD')
+      :pass
+    end
+  end
+end


config/default.yml
```diff
@@ -262,3 +262,14 @@ CommitMsg:
...
   TrailingPeriod:
     description: 'Checking for trailing periods in subject'
+
+PostCommit:
+  ALL:
+    requires_files: false
+    required: false
+    quiet: false
+
+  GitGuilt:
```
sds: This is a cute tool, but I don't see the functionality it provides as something that most developers would find useful to have by default. It doesn't speed up any particular process (that I'm aware of) and appears to be a vanity metric for the most part (feel free to give some concrete examples of where this would be useful in the general case).

Please add `enabled: false` so this does not run by default.

lencioni: git guilt might help you identify good people to review your commit.

lib/overcommit/hook/post_commit/git_guilt.rb
```diff
@@ -0,0 +1,9 @@
+module Overcommit::Hook::PostCommit
+  # Calculates the change in blame since the last revision.
+  class GitGuilt < Base
+    def run
+      system(executable, 'HEAD~', 'HEAD')
```
sds: Please use the `execute` helper for this, as it allows overcommit to properly handle setting up an appropriate environment for the child process to be spawned within.

```ruby
execute([executable, 'HEAD~', 'HEAD'])
```

sds:
Thanks for the pull request, @jawshooah.

Happy to merge once my comments are addressed!

lib/overcommit/hook/post_commit/git_guilt.rb
jawshooah: I originally wrote it this way:
```ruby
def run
  result = execute([executable, 'HEAD~', 'HEAD'])
  puts result.stdout
  :pass
end
```
But `execute` seems to strip all formatting (and newlines, weirdly) from `result.stdout`. This was the only way I could get the correctly formatted and colored output.

config/default.yml
jawshooah: @lencioni That's largely how I use it. It's good to know whose work you're clobbering. But I agree, this shouldn't be a default.

lib/overcommit/hook/post_commit/git_guilt.rb
sds: It sounds like `git guilt` only formats its output with newlines and color escape sequences if its writing to a TTY. Since in this case we're intercepting the output, it doesn't output color codes.

It's unfortunate it doesn't seem to have a `--color` flag to force color output, which is the standard solution to this sort of problem.

I'd rather not make an exception to using `execute` just for this feature. It should be easy enough to parse the output inline and manually output colors on its behalf. I realize Overcommit doesn't provide nice helpers for this, so for now you can just output the escape codes manually and we'll look into providing some helpers later.

sds:
Closing due to lack of response. Feel free to submit your pull request when it's ready.

PR #100, opened 2015-02-25
Add pre-commit hooks for w3c_validators

jawshooah:
Add pre-commit hooks for W3C's HTML and CSS validation services, accessed through the [w3c_validators](https://github.com/alexdunae/w3c_validators) gem.

`W3cCssValidator` is disabled by default, as the service seems a little finicky with large inputs.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -294,6 +294,7 @@ PreCommit:
       - '**/*.css'
 
   W3cHtmlValidator:
+    enabled: false
     description: 'Analyzing with W3C HTML validation service'
     validator_uri: 'http://validator.w3.org/check'
     charset: 'utf-8'
--- a/config/default.yml
+++ b/config/default.yml
@@ -283,6 +283,24 @@ PreCommit:
     description: 'Checking YAML syntax'
     include: '**/*.yml'
 
+  W3cCssValidator:
+    enabled: false
+    description: 'Analyzing with W3C CSS validation service'
+    validator_uri: 'http://jigsaw.w3.org/css-validator/validator'
+    language: 'en'
+    profile: 'css3'
+    warn_level: 2
+    include:
+      - '**/*.css'
+
+  W3cHtmlValidator:
+    description: 'Analyzing with W3C HTML validation service'
+    validator_uri: 'http://validator.w3.org/check'
+    charset: 'utf-8'
+    doctype: 'HTML5'
+    include:
+      - '**/*.html'
+
 # Hooks that run after HEAD changes or a file is explicitly checked out.
 PostCheckout:
   ALL:
--- a/lib/overcommit/hook/pre_commit/w3c_css_validator.rb
+++ b/lib/overcommit/hook/pre_commit/w3c_css_validator.rb
@@ -1,0 +1,74 @@
+module Overcommit::Hook::PreCommit
+  # Runs `w3c_validators` against any modified CSS files.
+  class W3cCssValidator < Base
+    def run
+      begin
+        require 'w3c_validators'
+      rescue LoadError
+        return :warn, 'w3c_validators not installed -- run `gem install w3c_validators`'
+      end
+
+      result_messages =
+        begin
+          applicable_files.collect do |path|
+            results = validator.validate_file(path)
+            messages = results.errors + results.warnings
+            messages.collect do |msg|
+              # Some warnings are not per-line, so use 0 as a default
+              line = Integer(msg.line || 0)
+
+              # Build message by hand to reduce noise from the validator response
+              text = "#{msg.type.to_s.upcase}; URI: #{path}; line #{line}: #{msg.message.strip}"
+              Overcommit::Hook::Message.new(msg.type, path, line, text)
+            end
+          end
+        rescue W3CValidators::ValidatorUnavailable => e
+          return :fail, e.message
+        rescue W3CValidators::ParsingError => e
+          return :fail, e.message
+        end
+
+      result_messages.flatten!
+      return :pass if result_messages.empty?
+
+      result_messages
+    end
+
+    private
+
+    def validator
+      unless @validator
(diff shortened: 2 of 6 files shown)


lib/overcommit/hook/pre_commit/w3c_css_validator.rb
```diff
@@ -0,0 +1,74 @@
+module Overcommit::Hook::PreCommit
+  # Runs `w3c_validators` against any modified CSS files.
+  class W3cCssValidator < Base
+    def run
+      begin
+        require 'w3c_validators'
+      rescue LoadError
+        return :warn, 'w3c_validators not installed -- run `gem install w3c_validators`'
```
sds: This should be a failure condition (i.e. return `:fail`). Any other hook that is missing its required executables currently fails.

lib/overcommit/hook/pre_commit/w3c_css_validator.rb
```diff
@@ -0,0 +1,74 @@
...
+            end
+          end
+        rescue W3CValidators::ValidatorUnavailable => e
+          return :fail, e.message
+        rescue W3CValidators::ParsingError => e
+          return :fail, e.message
+        end
+
+      result_messages.flatten!
+      return :pass if result_messages.empty?
```
sds: This case is handled for you automatically when you use `Overcommit::Hook::Messages` (see the `MessageProcessor#basic_status_and_output`)

lib/overcommit/hook/pre_commit/w3c_html_validator.rb
```diff
@@ -0,0 +1,71 @@
...
+            end
+          end
+        rescue W3CValidators::ValidatorUnavailable => e
+          return :fail, e.message
+        rescue W3CValidators::ParsingError => e
+          return :fail, e.message
+        end
+
+      result_messages.flatten!
+      return :pass if result_messages.empty?
```
sds: Ditto

lib/overcommit/hook/pre_commit/w3c_html_validator.rb
```diff
@@ -0,0 +1,71 @@
+module Overcommit::Hook::PreCommit
+  # Runs `w3c_validators` against any modified HTML files.
+  class W3cHtmlValidator < Base
+    def run
+      begin
+        require 'w3c_validators'
+      rescue LoadError
+        return :warn, 'w3c_validators not installed -- run `gem install w3c_validators`'
```
sds: Ditto

overcommit.gemspec
```diff
@@ -28,6 +28,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency             'childprocess', '~> 0.5.0'
 
+  s.add_development_dependency 'w3c_validators', '~> 1.2'
```
sds: Can you sort this?

config/default.yml
```diff
@@ -283,6 +283,24 @@ PreCommit:
...
+    enabled: false
+    description: 'Analyzing with W3C CSS validation service'
+    validator_uri: 'http://jigsaw.w3.org/css-validator/validator'
+    language: 'en'
+    profile: 'css3'
+    warn_level: 2
+    include:
+      - '**/*.css'
+
+  W3cHtmlValidator:
```
sds: This should also be opt-in (`enabled: false`). It's pretty heavyweight since it makes a remote request, and some organizations would have policies on sending private code to external entities.

sds:
Thanks for your contributions @jawshooah, and doubly so for writing thorough tests. 

Overall this looks good. Address those comments and I'll happily merge these hooks. Thanks!

lib/overcommit/hook/pre_commit/w3c_css_validator.rb
jawshooah: I figured as much, though `image_optim` warns if the gem isn't installed, and I used that as a base. I'll make the change.

lib/overcommit/hook/pre_commit/w3c_css_validator.rb
jawshooah: Not necessarily:
```ruby
def process_hook_return_value(hook_return_value)
  if hook_return_value.is_a?(Array) &&
     hook_return_value.first.is_a?(Message)
    # Process messages into a status and output
    Overcommit::MessageProcessor.new(
      self,
      @config['problem_on_unmodified_line'],
    ).hook_result(hook_return_value)
  else
    # Otherwise return as-is
    hook_return_value
  end
end
```
This only uses `Overcommit::MessageProcessor` if `hook_return_value` is a _non-empty_ `Array` of `Overcommit::Hook::Message` objects. If it's empty, it's returned as-is (i.e. `[]` rather than `[:pass]`).

config/default.yml
jawshooah: Done (diff isn't showing as outdated since the changed line is after this one)

lib/overcommit/hook/pre_commit/w3c_css_validator.rb
sds: Hmm, well that wasn't the intent. :)

Feel free to leave as-is then. I'll circle back on this at some point. An empty array can safely be assumed to be an empty list of `Overcommit::Hook::Message`s.

sds:
Thanks for addressing those comments, @jawshooah.

Merged in 4a95f97 with some RuboCop lint fixes in 56c2786 (Travis CI is having problems right now, so you didn't get these errors yourself).

I also removed the `Validator` portion of the name from the hooks, as it seemed redundant.

Thanks again!

lib/overcommit/hook/pre_commit/w3c_css_validator.rb
sds: Fixed treatment of empty arrays in e21d6d8ce5a2715e16464369d267d2d9d5784bcc.

PR #101, opened 2015-02-25
Fail image_optim hook if gem not installed

jawshooah:
Per discussion in #100.


sds:
Merged in b16ce1a8512986491b05cd093ee6c16bf8145cf5. Thanks!

PR #103, opened 2015-02-25
Quote `file_path` in `extract_modified_lines`

jawshooah:
This fixes a bug where lines modified in files with spaces in their
paths are not recognized.


sds:
Thanks! Merged in 5e10d1e.

PR #104, opened 2015-02-25
Add pre-commit hook for checkstyle

jawshooah:
Add hook for [Checkstyle](http://checkstyle.sourceforge.net), a style checker for Java.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -111,6 +111,12 @@ PreCommit:
       - 'config/settings.yml'
       - 'config/settings/**/*.yml'
 
+  Checkstyle:
+    description: 'Analyzing with checkstyle'
+    required_executable: 'checkstyle'
+    flags: ['-c', '/sun_checks.xml']
+    include: '**/*.java'
+
   CoffeeLint:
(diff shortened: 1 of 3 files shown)


config/default.yml
```diff
@@ -111,6 +111,12 @@ PreCommit:
       - 'config/settings.yml'
       - 'config/settings/**/*.yml'
 
+  Checkstyle:
+    description: 'Analyzing with checkstyle'
+    required_executable: 'checkstyle'
+    flags: ['-c', '/sun_checks.xml']
```
sds: I assume this file path is standard? (weird that it has a leading slash)

sds:
Looks good. Merged in 75226de4f00a592cff591f09a4b580b2e2967264 and renamed to `JavaCheckstyle` in 4da8aa775e775416185d402f0925b4c6244f9a52, as the former was a bit too vague.

config/default.yml
jawshooah: Weird indeed, but that's the standard config. The leading slash just indicates that it's built-in, I assume.

PR #106, opened 2015-02-26
Add pre-commit hook for pep8

jawshooah:
Add hook for [pep8](https://pypi.python.org/pypi/pep8), a Python linter. Disable by default, as [flake8](https://pypi.python.org/pypi/flake8) already includes pep8.

This is useful for those who prefer to omit warnings from [pyflakes](https://pypi.python.org/pypi/pyflakes), which is also included in flake8.


sds:
Merged in c020fc7.

PR #107, opened 2015-02-26
Add pre-commit hook for pyflakes

jawshooah:
Add hook for [pyflakes](https://pypi.python.org/pypi/pyflakes). Disable by default, as [flake8](https://pypi.python.org/pypi/flake8) includes pyflakes.

This is useful for those who want semantic linting without the style checking of [pep8](https://pypi.python.org/pypi/pep8).

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -199,6 +199,13 @@ PreCommit:
       - '**/*.rb'
       - '**/*.rake'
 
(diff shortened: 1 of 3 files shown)


sds:
Merged in 87c0e6831c14e52ca539540195b35d4549e02c6f.

PR #108, opened 2015-02-26
Add pre-commit hook for pep257

jawshooah:
Add hook for [pep257](https://pypi.python.org/pypi/pep257), a linter for Python docstrings.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -198,6 +198,12 @@ PreCommit:
     required_executable: 'grep'
     flags: ['-IHn', '^<<<<<<<\s']
 
+  Pep257:
+    description: 'Analyzing with pep257'
+    required_executable: 'pep257'
+    install_command: 'pip install pep257'
+    include: '**/*.py'
+
   Pep8:
     enabled: false
     description: 'Analyzing with pep8'
--- a/config/default.yml
+++ b/config/default.yml
@@ -199,7 +199,7 @@ PreCommit:
     flags: ['-IHn', '^<<<<<<<\s']
 
   Pep257:
-    description: 'Analyzing with pep257'
+    description: 'Analyzing docstrings with pep257'
     required_executable: 'pep257'
     install_command: 'pip install pep257'
     include: '**/*.py'
--- a/lib/overcommit/hook/pre_commit/pep257.rb
+++ b/lib/overcommit/hook/pre_commit/pep257.rb
@@ -1,0 +1,19 @@
+module Overcommit::Hook::PreCommit
+  # Runs `pep257` against any modified Python files.
+  class Pep257 < Base
+    def run
+      result = execute(command + applicable_files)
+      return :pass if result.success?
+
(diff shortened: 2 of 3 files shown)


config/default.yml
```diff
@@ -198,6 +198,12 @@ PreCommit:
     required_executable: 'grep'
     flags: ['-IHn', '^<<<<<<<\s']
 
+  Pep257:
+    description: 'Analyzing with pep257'
```
sds: Since PEP257 specifically deals with docstrings, we should probably say that here.

```
Analyzing docstrings with pep257`
```

config/default.yml
```diff
@@ -198,6 +198,12 @@ PreCommit:
     required_executable: 'grep'
     flags: ['-IHn', '^<<<<<<<\s']
 
+  Pep257:
```
sds: I'm unsure about making this run by default (having never heard about this PEP until now).

Is this something people actually follow? What sort of adoption is there in the greater Python community?

jawshooah: [PEP](https://www.python.org/dev/peps/) is the official channel for Python Enhancement Proposals, which are very much known and followed throughout the Python community.

[PEP 8](https://www.python.org/dev/peps/pep-0008/) and [PEP 257](https://www.python.org/dev/peps/pep-0257) together comprise the official Python [style guide](https://www.python.org/doc/essays/styleguide/) conceived by the language's BDFL, Guido van Rossum.

So yes, I'd wager most Python devs are aware :wink:

sds: Sorry, I specifically meant PEP257 (I'm familiar with PEP in general).

However, having dug into the spec this makes sense as a default.

sds:
Merged in 8dc7af28e4b3ac2011c325e1385373e31ed63692.

sds:
You about done for the near future, @jawshooah?

I'd like to cut a new release of Overcommit, but am willing to wait if you had a few more hooks you wanted to contribute (I'd rather not cut two releases in quick succession). 

Let me know, otherwise I'll cut the next release tomorrow (it includes some helpful bug fixes).

config/default.yml
jawshooah: Agreed, though now I'm second-guessing myself as to whether a failed check should return `:fail`. It might make more sense to return `:warn` and let the user opt-in to failure with `on_warn: 'fail'`.

As it is, there are several options if someone doesn't care much about docstring conventions. They can:
- Disable the hook entirely
- Set `on_fail: 'warn'`
- Ignore specific errors with `flags: ['--ignore=D101,D202']`, etc

jawshooah:
Just about. I'd also like to improve `PythonFlake8`, which currently doesn't do any meaningful parsing of error messages.

config/default.yml
sds: I'm comfortable with the idea of making this a hard failure, with users having the choice to opt-out. The PEP specification could use some help with adoption.

PR #109, opened 2015-02-26
Improve message parsing for PythonFlake8

jawshooah:
Use specific rules to distinguish errors from warnings.

`ERROR_REGEX` taken from https://github.com/SublimeLinter/SublimeLinter-flake8.  The file in question has a copyright notice:

```python
#
# linter.py
# Linter for SublimeLinter3, a code checking framework for Sublime Text 3
#
# Written by Aparajita Fishman
# Copyright (c) 2013 Aparajita Fishman
#
# License: MIT
#
```

How do I properly attribute this in code?

### DIFF
--- a/lib/overcommit/hook/pre_commit/python_flake8.rb
+++ b/lib/overcommit/hook/pre_commit/python_flake8.rb
@@ -1,29 +1,13 @@
 module Overcommit::Hook::PreCommit
   # Runs `flake8` against any modified Python files.
   class PythonFlake8 < Base
-    # The following regex marks these pyflakes and pep8 codes as errors.
-    # All other codes are marked as warnings.
-    #
-    # Pyflake Errors:
-    #  - F402 import module from line N shadowed by loop variable
-    #  - F404 future import(s) name after other statements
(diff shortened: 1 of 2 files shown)


jawshooah:
For reference, the list of errors was specified in SublimeLinter/SublimeLinter-flake8#6.

jawshooah:
If we'd prefer to steer clear of such subjective distinctions, we could simply parse `E` and `F` messages as errors, and `W`, `C`, `N` as warnings.

sds:
I'm not familiar enough with these errors to know how useful/annoying they are, but I would prefer we rely on the tool to indicate error/warning rather than Overcommit making that decision. This way the conversation occurs on the tool itself, not in Overcommit.

Otherwise looks good.

jawshooah:
In that case, I'll go by the `flake8` [documentation](http://flake8.readthedocs.org/en/latest/warnings.html) and parse `Exxx`, `Fxxx` as errors and everything else as warnings.

sds:
Merged in 3342a22.

PR #110, opened 2015-02-26
Add pre-commit hook for tidy

jawshooah:
Add hook for [tidy](http://www.html-tidy.org/), an HTML linter (among other things).

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -154,6 +154,12 @@ PreCommit:
       - '**/Makefile'
       - '.gitmodules'
 
+  HtmlTidy:
+    description: 'Analyzing HTML with tidy'
+    required_executable: 'tidy'
+    flags: ['-errors', '-quiet', '-utf8']
+    include: '**/*.html'
+
   ImageOptim:
     description: 'Checking for optimizable images'
     include:
--- a/config/default.yml
+++ b/config/default.yml
@@ -155,6 +155,7 @@ PreCommit:
       - '.gitmodules'
 
   HtmlTidy:
+    enabled: false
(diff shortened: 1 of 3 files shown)


config/default.yml
```diff
@@ -154,6 +154,12 @@ PreCommit:
       - '**/Makefile'
       - '.gitmodules'
 
+  HtmlTidy:
```
sds: What sort of use does this tool have, and how does it relate to the `W3cHtml` pre-commit hook?

I'm tempted say this should be opt-in.

jawshooah: In general, tidy [identifies and corrects markup errors](http://www.html-tidy.org/documentation/). Given the `-errors` flag, it foregoes the corrections and just outputs the errors.

It's not related to `W3cHtml`, though their rulesets likely overlap. It also doesn't make any remote requests, which is a plus.

sds:
Sorry, by "use" I meant developer adoption.

After digging into this tool a bit, it does indeed look interesting. However, it looks like you need to be running a very recent version of it to get some benefits (like HTML5 support). 

Given that most operating systems ship with an old version by default (my Mac comes with a version of from 2006, and my CentOS Linux machine comes with one from 2009 after I install via `yum`), it should probably be an opt-in as it seems like a pretty high bar to require people to compile code on their system to start using a package required by a hook.

jawshooah:
Fair enough! I've pushed the change.

For your own enjoyment, you can install the HTML5 version with `brew install --head tidy` :smile:

sds:
Merged in 0f8d160.

PR #111, opened 2015-02-27
Add pre-commit hook for Pylint

jawshooah:
Add hook for [Pylint](http://www.pylint.org/), a static analyzer for Python.

Last one for a while, I swear!


sds:
Merged in 6583e01

PR #112, opened 2015-02-27
Parse errors and warnings in csslint output

jawshooah:
Add logic to distinguish between errors and warnings in csslint output.


sds:
Merged in 852e689

PR #113, opened 2015-02-27
Remove :modified_lines_in_file stubs in pre-commit hook specs

jawshooah:
Pre-commit hook specs never use this method, so these lines just create cruft and could be misleading to someone reading or modifying the specs (read: myself).

To clarify, the hook specs directly check the output of `run` without any processing to filter out unmodified lines.


sds:
Good catch. It looks like the refactoring I did in 89b7d7b4 to add the `extract_messages` pattern removed the need for this stubbing and I forgot to kill it.

Merged in c87bca701ce887cd74e1d0729b801181fe847796 with some additional cleanup in 6594a260c8951571f4e521130eadf30b5192d359.

PR #114, opened 2015-02-28
LocalPathsInGemfile ignores commented local paths

nevinera:
We use a local path for development constantly, and leave it commented in the Gemfile. It seems appropriate to me that the git hook shouldn't mind finding a local path in a gemfile if it's in a comment.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -189,7 +189,7 @@ PreCommit:
   LocalPathsInGemfile:
     description: 'Checking for local paths in Gemfile'
     required_executable: 'grep'
-    flags: ['-IHnE', '(\bpath:)|(:path\s*=>)']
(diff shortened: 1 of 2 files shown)


sds:
Sure, that sounds fine. Why you're committing commented lines of code is another question, but the premise for this enhancement is reasonable.

Merged in 1a61fa7.

nevinera:
These days it's for convenience, but the original and more justifiable comment was a description of how to do development on the main application with a local copy of a private gem - we had a comment like:

```ruby
# To work with a local version of the engine, use this line instead:
# gem 'engine', '= 4.3.1', :path => '../engine', :require => 'engine-clean'
gem 'engine', '= 4.3.1', :require => 'engine-clean'
```

sds:
Ah, makes sense. Thanks for explaining.

PR #115, opened 2015-03-02
Ignore blank lines in csslint output

jawshooah:
csslint outputs errors as blocks for each file, with a blank line separating them. Including these lines in the call to `extract_messages` results in the error:
```
Unexpected output: unable to determine line number or type of error/warning for message
```


sds:
Merged in c977e5a.

PR #116, opened 2015-03-02
Allow extraction of messages with no line number

jawshooah:
Some linters output file-level errors that don't correspond to a
specific line number. Previously, this was only workable by either
manually building the Overcommit::Hook::Message instance or
splicing a dummy line number into the output.


sds:
Merged in 908d0db960417d641b5a07f91b11ffa2f883de63.

PR #118, opened 2015-03-03
Extract file, line, type from coffeelint output

jawshooah:
Use `raw` JSON reporter to extract information from coffeelint messages. There is no one-line reporter, so this is the best we can do. There is a `csv` reporter, but it's not as safe since syntax error messages include newlines and unicode.


sds:
Merged in 5785900 and 923be23. Thanks!

PR #119, opened 2015-03-03
Ignore hard tabs in Godeps.json

sectioneight:
Godep is a utility for managing Go dependencies. It uses the default
JSON marshaller to automatically save this file (should never be
hand-edited). We should not error out or this, since we have no control
over it (and even if we un-tabbed it, it would be re-tabbed next time we
change our dependencies).

Here is a sample error I get:
```
$ git commit -m "bump sqlite3"
Running pre-commit hooks
Checking for hard tabs.....................................[HardTabs] FAILED
Hard tabs detected:
/Users/aiden/src/gocode/src/uber/uberalls/Godeps/Godeps.json:2: "ImportPath": "uber/uberalls",
/Users/aiden/src/gocode/src/uber/uberalls/Godeps/Godeps.json:3: "GoVersion": "go1.4",
/Users/aiden/src/gocode/src/uber/uberalls/Godeps/Godeps.json:4: "Deps": [

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -151,6 +151,7 @@ PreCommit:
     flags: ['-IHn', "\t"]
     exclude:
       - '**/*.go'
+      - 'Godeps/Godeps.json'
       - '**/Makefile'
       - '.gitmodules'
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -151,9 +151,9 @@ PreCommit:
     flags: ['-IHn', "\t"]
     exclude:
       - '**/*.go'
-      - 'Godeps/Godeps.json'
       - '**/Makefile'
       - '.gitmodules'
+      - 'Godeps/Godeps.json'
 
   HtmlTidy:
     enabled: false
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: config/default.yml [hunk 1]
```diff
@@ -151,6 +151,7 @@ PreCommit:
     flags: ['-IHn', "\t"]
     exclude:
       - '**/*.go'
+      - 'Godeps/Godeps.json'
```
sectioneight: Not sure about the preferred alphabetization here. I grouped it with the previous `.go` ignore for logical grouping but maybe we're going for alphabetization here?

lencioni: We seem to have a bit of a mix in this file, and I don't think it matters much, but I would personally go with alphabetization.

sectioneight:
Also: do I need to configure something to access gerrit? I tried to go to upload this there first but can't connect. I'm assuming it is now behind a VPN; let me know if I'm wrong.

lencioni:
I merged this in e6095c69b7bcfbfc7dd92fb7822efcf6f006f367 (without your alphabetization, sorry!). Thanks for the PR!

Correct, Gerrit is now fully behind a VPN so you'll need to just use GitHub for this stuff going forward.

sectioneight:
Thanks, Joe!

PR #120, opened 2015-03-03
Extract file, line from golint output

jawshooah:
This and #118 cover the last linter hooks whose message info is not extracted from the output.


sds:
Merged in 29f319e. Thanks!

Issue #117, opened 2015-03-02
`ALL` hook settings in config are overridden by defaults


jawshooah:
Currently, any per-hook settings in `default.yml` override those under `ALL` in the user's `.overcommit.yml`.

For example, say I want to exclude a set of files in my repository for all pre-commit hooks. In my `.overcommit.yml`:

```yaml
PreCommit:
  ALL:
    exclude:
      - 'vendor/**/*'
```

These files may be auto-generated or externally developed (and likely shouldn't be in the repository at all, but that's a separate concern), and may not be assumed to follow the style conventions set forth for the rest of the repository.

However, some hooks already exclude certain types of files by default. For example:

```yaml
HardTabs:
  exclude:
    - '**/*.go'
    - '**/Makefile'
    - '.gitmodules'

TrailingWhitespace:
  exclude:
    - '**/db/structure.sql'
```

These exclusions override those in my settings for all pre-commit hooks, so I have to explicitly override them on a per-hook basis.

This isn't necessarily bad behavior, but I figured it was worth discussing, as it requires the user to know which hooks already have default exclusions in order to override them.

sds:
Yeah, we've run into this as well. The behavior was changed in 483a306 to fix an issue where you could not override array values (it would always merge with the previous value).

I'm not sure if there's a solution here that makes sense while still allowing you the ability to override configuration when you need to. Perhaps we simply need to do a better job of documenting in what order configuration is loaded.

I'm open to suggestions on this if you have any.

jawshooah:
I'm tempted to say that any settings in the user config, including `ALL`, should override those in the default. So when building the config object for a hook, the merge order would be:

default (ALL) -> default (hook) -> user (ALL) -> user (hook)

I haven't gone through the configuration code yet, so I'm not sure whether this would be trivial to implement. I just feel that it matches user expectations (or at least mine) better than the current behavior.

lencioni:
> `default (ALL) -> default (hook) -> user (ALL) -> user (hook)`

I think that may have been the previous behavior, but it has the problem of not allowing people to override array values from default configuration. Right?

jawshooah:
Not sure why that would be the case. Why should array values be treated any differently than other settings?

sds:
Implemented by #121. Thanks!

PR #121, opened 2015-03-03
Override default settings with user 'ALL' hook

jawshooah:
Settings specified in the user config file under 'ALL' should
override default hook-specific settings.

Ref #117

### DIFF
--- a/lib/overcommit/configuration.rb
+++ b/lib/overcommit/configuration.rb
@@ -168,6 +168,10 @@ module Overcommit
     end
 
     def smart_merge(parent, child)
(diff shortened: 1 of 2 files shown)


jawshooah:
@sds Any thoughts on this, or my recent linter-specific PRs?

sds:
I'm usually not a fan of these kinds of special cases in low-level helper code, but I appreciate the fact that it works and is relatively isolated.

I also agree that this behavior is more intuitive.

Merged in b0e0e6f with some additional specs in 863836a. Thanks!

PR #123, opened 2015-03-05
Add pre-commit hook for ruby-lint

jawshooah:
https://github.com/YorickPeterse/ruby-lint

This linter is still a bit rough around the edges, so disable it by default.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -285,6 +285,19 @@ PreCommit:
       - '**/Gemfile'
       - '**/Rakefile'
 
+  RubyLint:
+    enabled: false
+    description: 'Analyzing with ruby-lint'
+    required_executable: 'ruby-lint'
+    flags: ['--presenter=syntastic', '--levels=error,warning']
+    install_command: 'gem install ruby-lint'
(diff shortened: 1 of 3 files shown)


jawshooah:
Is that a commonly used extension for Ruby files? It isn't included in the Rubocop or Reek hooks.

lencioni:
It is the extension of the config file used by rack-based applications, such as Rails. Probably 99% of the time it will be `config.ru`, so maybe that would be better?

https://github.com/rack/rack/wiki/%28tutorial%29-rackup-howto

jawshooah:
Ah, I see. That change should probably be made in a separate PR, so it can be applied to Reek and Rubocop as well.

jawshooah:
Actually, I need to edit the inclusions for this linter in particular, since it's intended to be used only on Ruby source files (i.e. not `Gemfile`, `Rakefile` or `*.gemspec`).

sds:
Merged in 8a8df547734fe854cd537f76148acd4499be8054. Thanks!

PR #124, opened 2015-03-08
Exclude Markdown files from TrailingWhitespace

jawshooah:
Two trailing spaces are used to force a line break in Markdown files.

Change-Id: I6aea9d7b3da14381dd2fcf7f80648a419ba70ac5

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -323,6 +323,8 @@ PreCommit:
     description: 'Checking for trailing whitespace'
     required_executable: 'grep'
     flags: ['-IHn', '\s$']
+    exclude:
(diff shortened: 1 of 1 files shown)


lencioni:
Thanks for the pull request. This has been merged in e4b5b33cabe133b75c846495f2584db9bdf31540.

sds:
Eventually we're going to want to draw some sort of line of what special exceptions Overcommit should support first-hand. We already have some stuff for Go and now Markdown--I'd like to not be a central repository filled with special cases, even if the configuration is simple.

In any case, this is fine.

PR #125, opened 2015-03-08
Add pre-commit hook for JSL (JavaScript Lint)

jawshooah:
[JavaScript Lint](http://www.javascriptlint.com) is a static analyzer in the vein of JSLint, JSHint, etc.

Unfortunately this tool is not available through a cross-platform package manager, so no `install_command`. The `javascriptlint` package on `npm` is unmaintained (no commits in 3 years) and does not provide an executable.

If you're on a Mac, though, you can get it with `brew install jsl` :smile:

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -192,6 +192,12 @@ PreCommit:
     install_command: 'npm install -g jshint'
     include: '**/*.js'
 
+  Jsl:
+    description: 'Analyzing with JSL'
+    required_executable: 'jsl'
+    flags: ['-nologo', '-nofilelisting', '-nocontext', '-nosummary']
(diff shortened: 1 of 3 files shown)


sds:
If there is no convenient way to install a tool via well-known package manager, I would argue such a hook should be disabled by default.

Merged in bf2366da9ae28d4beab316983c59aca88c573e4e. Disabled by default in b5c79ff9d3538039e0c74d9bfa2b4d1dbacf05ed. 

Thanks!

jawshooah:
Ah, for some reason I thought that if `required_executable` wasn't in the path, the hook simply didn't run. Thanks for the merge and the fix!

PR #130, opened 2015-03-12
Add GoVet pre-commit hook

sectioneight:
From the docs:

    https://godoc.org/golang.org/x/tools/cmd/vet

    Vet examines Go source code and reports suspicious constructs, such
    as Printf calls whose arguments do not align with the format
    string. Vet uses heuristics that do not guarantee all reports are
    genuine problems, but it can find errors not caught by the
    compilers.

Note some shennanigans around detecting whether the tool is installed. When you run `go get golang.org/x/tools/cmd/vet` it does not install a binary in your path; rather you invoke it with `go tool vet` which will exit with an error message if the utility is not installed. I am not sure if there is prior art in this repo for doing this in a better way; please feel free to suggest one!

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -142,6 +142,13 @@ PreCommit:
     install_command: 'go get github.com/golang/lint/golint'
     include: '**/*.go'
 
+  GoVet:
+    description: 'Analyzing with go vet'
+    required_executable: 'go'
+    flags: ['tool', 'vet']
+    install_command: 'go get golang.org/x/tools/cmd/vet'
+    include: '**/*.go'
+
   HamlLint:
     description: 'Analyzing with haml-lint'
     required_executable: 'haml-lint'
--- a/lib/overcommit/hook/pre_commit/go_vet.rb
+++ b/lib/overcommit/hook/pre_commit/go_vet.rb
@@ -1,0 +1,20 @@
+module Overcommit::Hook::PreCommit
+  # Runs `go vet` against any modified Golang files.
+  class GoVet < Base
+    def run
+      result = execute(command + applicable_files)
+      return :pass if result.success?
+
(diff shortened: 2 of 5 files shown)


lib/overcommit/hook/pre_commit/go_vet.rb
```diff
@@ -0,0 +1,20 @@
+module Overcommit::Hook::PreCommit
+  # Runs `go vet` against any modified Golang files.
+  class GoVet < Base
+    def run
+      result = execute(command + applicable_files)
+      return :pass if result.success?
+
+      if result.stderr =~ /no such tool "vet"/
+        return :warn, "`go tool vet` is not installed#{install_command_prompt}"
```
sds: We've recently changed the preferred failure mode for dependencies to actually `:fail` instead of `:warn`. If you have the hook enabled, it should fail unless it actually ran successfully.

config/default.yml
```diff
@@ -142,6 +142,13 @@ PreCommit:
     install_command: 'go get github.com/golang/lint/golint'
     include: '**/*.go'
 
+  GoVet:
+    description: 'Analyzing with go vet'
+    command: 'go'
```
sds: Typically we use `required_executable` for this.

sds:
Yeah, we don't have a way to specify that a package under a particular toolchain exists, only that an executable itself exists. I'm not sure this is something for which a general catch all solution exists, but I'm open to one if you're interested in adding one.

Otherwise, having the hook fail when `vet` isn't available would be the preferred behavior.

sds:
Merged in 8de6506158a9f0dffcf56fd1f4807cd3d03bcfaa. Thanks!

sectioneight:
wheee, thanks shane

PR #131, opened 2015-03-12
Add simplecov coverage tool to spec runs

sectioneight:
I'm out of the Ruby game, so this may be the wrong tool, but when
contributing code I like to know that my change is well-tested. I tried
simplecov and it worked nicely. Perhaps this would be useful for others?

Example report: http://sca.ndella.com/overcommit-coverage/index.html#_AllFiles

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 Gemfile.lock
 pkg/
 .bundle
+coverage
--- a/overcommit.gemspec
+++ b/overcommit.gemspec
@@ -31,4 +31,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rspec', '~> 3.0'
   s.add_development_dependency 'travis', '~> 1.7'
   s.add_development_dependency 'w3c_validators', '~> 1.2'
+  s.add_development_dependency 'simplecov', '~> 0.9.0'
 end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,6 +1,9 @@
 require 'overcommit'
 require 'tempfile'
 
+require 'simplecov'
+SimpleCov.start
+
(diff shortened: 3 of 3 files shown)


lencioni:
We are already using code climate, so it would be nice to hook coverage into that service (and add a badge to the README).

- https://codeclimate.com/tour/test-coverage
- http://docs.codeclimate.com/article/219-setting-up-test-coverage

I think maybe @sds needs to do this.

sectioneight:
Yea I'm not super familiar with the toolchain, but if there's a way we can surface coverage information to developers before stuff gets merged into master, it may help with contributing better code.

lencioni:
Yeah, that's a great point. I think that Code Climate won't comment on PRs, but Coveralls will. https://coveralls.io

sds:
I ended up adding a Coveralls badge in bbf349a9bce00131236219f5639d79444e4fe719.

Usually not a fan of LOC-based coverage stats, but I've set it up to alert if there's a major drop in testing.

Thanks for making the suggestion.

lencioni:
> Usually not a fan of LOC-based coverage stats

Why not?

sds:
It encourages a type of testing that gives you a false sense of security.

Consider the following code:

```ruby
if condition
  puts 'Huzzah'
end
```

With LOC-based testing, I would only need to have a test that exercises `condition` evaluating to `true` in order to get 100% coverage. What never ends up getting tested is the situation where `condition` is `false`.

[Branch coverage](http://en.wikipedia.org/wiki/Code_coverage#Basic_coverage_criteria) (where every edge of the code path is traversed), is a stronger set of criteria for code coverage, but even it isn't perfect. There is also path coverage and prime path coverage which are mostly of interest in academia or when building an airplane. :smile:

PR #132, opened 2015-03-18
Add pre-commit hook to check for case conflicts

jawshooah:
Files with the same name but different capitalization can cause
trouble on case-insensitive filesystems.

### DIFF
--- a/lib/overcommit/git_repo.rb
+++ b/lib/overcommit/git_repo.rb
@@ -54,6 +54,19 @@ module Overcommit
         map { |relative_file| File.expand_path(relative_file) }
     end
 
+    # Returns the names of files in the given paths that are tracked by git.
+    #
+    # @param paths [Array<String>] list of paths to check
+    # @option options [String] ref ('HEAD') Git ref to check
+    # @return [Array<String>] list of absolute file paths
+    def list_files(paths = [], options = {})
+      ref = options[:ref] || 'HEAD'
+      `git ls-tree --name-only #{ref} #{paths.join(' ')}`.
+        split(/\n/).
+        map { |relative_file| File.expand_path(relative_file) }.
+        reject { |file| File.directory?(file) } # Exclude submodule directories
+    end
+
     # Returns the names of all files that are tracked by git.
     #
     # @return [Array<String>] list of absolute file paths
--- a/lib/overcommit/hook/pre_commit/case_conflicts.rb
+++ b/lib/overcommit/hook/pre_commit/case_conflicts.rb
@@ -8,7 +8,7 @@ module Overcommit::Hook::PreCommit
       conflict_hash = repo_files.classify(&:downcase).
         select { |_, files| files.size > 1 }
       conflict_files = applicable_files.
-        select { |file| conflict_hash.include? file.downcase }
+        select { |file| conflict_hash.include?(file.downcase) }
 
       conflict_files.map do |file|
         conflicts = conflict_hash[file.downcase].map { |f| File.basename(f) }
--- a/lib/overcommit/hook/pre_commit/case_conflicts.rb
+++ b/lib/overcommit/hook/pre_commit/case_conflicts.rb
@@ -3,7 +3,8 @@ module Overcommit::Hook::PreCommit
   # Adapted from https://github.com/pre-commit/pre-commit-hooks
   class CaseConflicts < Base
     def run
-      repo_files = Set.new(Overcommit::GitRepo.all_files)
+      paths = Set.new(applicable_files.map { |file| File.dirname(file) })
+      repo_files = Set.new(Overcommit::GitRepo.list_files(paths.to_a) + applicable_files)
       conflict_hash = repo_files.classify(&:downcase).
         select { |_, files| files.size > 1 }
       conflict_files = applicable_files.
--- a/lib/overcommit/hook/pre_commit/case_conflicts.rb
+++ b/lib/overcommit/hook/pre_commit/case_conflicts.rb
@@ -1,0 +1,19 @@
+module Overcommit::Hook::PreCommit
+  # Checks for files that would conflict in case-insensitive filesystems
+  # Adapted from https://github.com/pre-commit/pre-commit-hooks
+  class CaseConflicts < Base
+    def run
+      repo_files = Set.new(Overcommit::GitRepo.all_files)
+      conflict_hash = repo_files.classify(&:downcase).
+        select { |_, files| files.size > 1 }
+      conflict_files = applicable_files.
+        select { |file| conflict_hash.include? file.downcase }
+
+      conflict_files.map do |file|
+        conflicts = conflict_hash[file.downcase].map { |f| File.basename(f) }
+        msg = "Case-insensitivity conflict detected: #{conflicts.join(', ')}"
+        Overcommit::Hook::Message.new(:error, file, nil, msg)
+      end
+    end
+  end
+end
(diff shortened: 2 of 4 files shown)


lib/overcommit/hook/pre_commit/case_conflicts.rb
```diff
@@ -0,0 +1,19 @@
+module Overcommit::Hook::PreCommit
+  # Checks for files that would conflict in case-insensitive filesystems
+  # Adapted from https://github.com/pre-commit/pre-commit-hooks
+  class CaseConflicts < Base
+    def run
+      repo_files = Set.new(Overcommit::GitRepo.all_files)
```
sds: I'm worried that this will be overly expensive on large repositories. Would it be possible to adjust it to only get files in the same directories as the files that changed?

lib/overcommit/hook/pre_commit/case_conflicts.rb
```diff
@@ -0,0 +1,19 @@
...
+      conflict_hash = repo_files.classify(&:downcase).
+        select { |_, files| files.size > 1 }
+      conflict_files = applicable_files.
+        select { |file| conflict_hash.include? file.downcase }
+
+      conflict_files.map do |file|
+        conflicts = conflict_hash[file.downcase].map { |f| File.basename(f) }
+        msg = "Case-insensitivity conflict detected: #{conflicts.join(', ')}"
```
sds: I would change this message to emphasize that this is a problem with case-insensitive filesystems.

```
Conflict detected for case-insensitive file systems: file1.txt, File1.txt
```

lib/overcommit/hook/pre_commit/case_conflicts.rb
jawshooah: Perhaps by adding a method to `GitRepo` to list the files in a particular directory or list of directories. Something like

```ruby
# Returns the names of files in the given paths that are tracked by git.
#
# @return [Array<String>] list of absolute file paths
def list_files(*paths)
  `git ls-files #{paths.join(' ')}`.
    split(/\n/).
    map { |relative_file| File.expand_path(relative_file) }.
    reject { |file| File.directory?(file) } # Exclude submodule directories
end

# Returns the names of all files that are tracked by git.
#
# @return [Array<String>] list of absolute file paths
def all_files
  list_files
end
```

jawshooah: Then in `case_conflicts.rb`:

```ruby
paths = Set.new(applicable_files.map{ |f| File.dirname(f) })
repo_files = Set.new(Overcommit::GitRepo.list_files(paths))
```

lib/overcommit/git_repo.rb
```diff
@@ -54,16 +54,24 @@ def modified_files(options)
...
     end
 
-    # Returns the names of all files that are tracked by git.
+    # Returns the names of files in the given paths that are tracked by git.
     #
+    # @param paths [Array<String>] variable-length list of paths to check
     # @return [Array<String>] list of absolute file paths
-    def all_files
-      `git ls-files`.
+    def list_files(*paths)
```
sds: I don't think there is a need to make this a splat? Instead it can just default to the empty array. As long as it's an `Enumerable` we don't need to care.

Having to remember whether a method takes a list of arguments or just a list, I would prefer it always took the latter.

lib/overcommit/git_repo.rb
```diff
@@ -54,16 +54,24 @@ def modified_files(options)
...
+      `git ls-files #{paths.join(' ')}`.
```
sds: Notice that `git ls-files` will print _everything_ under the directory, so we're still in the same situation as before if we edit a file at the root level of the repository.

Shy of finding a flag that disables this recursive behavior, we'll need to only return the files that are in the directories. Better than returning a set of all files in the repository.

jawshooah: Looks like `git ls-tree HEAD --name-only <paths>` is what we want. I'll make the change.

lib/overcommit/git_repo.rb
jawshooah: Agreed, changed in my last push.

lib/overcommit/hook/pre_commit/case_conflicts.rb
```diff
@@ -0,0 +1,20 @@
+module Overcommit::Hook::PreCommit
+  # Checks for files that would conflict in case-insensitive filesystems
+  # Adapted from https://github.com/pre-commit/pre-commit-hooks
+  class CaseConflicts < Base
+    def run
+      paths = Set.new(applicable_files.map { |file| File.dirname(file) })
```
sds: `File.dirname` returns the directory without a trailing slash. `git ls-tree` only returns all files in the directory if the directory name has a trailing `/` (I'm running git 1.8.3.1). Did you test this?

jawshooah: Ach, my bad. I was adding the trailing slash without thinking when testing `git ls-tree`. Fix incoming.

lib/overcommit/hook/pre_commit/case_conflicts.rb
```diff
@@ -0,0 +1,20 @@
+module Overcommit::Hook::PreCommit
+  # Checks for files that would conflict in case-insensitive filesystems
+  # Adapted from https://github.com/pre-commit/pre-commit-hooks
+  class CaseConflicts < Base
+    def run
+      paths = Set.new(applicable_files.map { |file| File.dirname(file) + File::SEPARATOR })
+      repo_files = Set.new(Overcommit::GitRepo.list_files(paths.to_a) + applicable_files)
```
sds: You shouldn't need `paths.to_a` here, since the set is `Enumerable`.

lib/overcommit/hook/pre_commit/case_conflicts.rb
```diff
@@ -0,0 +1,20 @@
...
+      conflict_hash = repo_files.classify(&:downcase).
+        select { |_, files| files.size > 1 }
+      conflict_files = applicable_files.
+        select { |file| conflict_hash.include? file.downcase }
```
sds: This codebase tries to subscribe to [Jim Weirich](https://en.wikipedia.org/wiki/Jim_Weirich)'s approach to parentheses for method calls, which is to include them [if you care about the return value](https://twitter.com/jimweirich/status/373814960788221952) (which we do in this case).

jawshooah: Cool beans, I'll make the change.

lib/overcommit/hook/pre_commit/case_conflicts.rb
jawshooah: Necessary in this case, as `join` is defined on `Array` but not `Enumerable`.

sds: Ah, I didn't realize. In hindsight this makes sense since a set doesn't have the concept of order of elements.

jawshooah: Yup. Converting to a `Set` may be too much overhead for large commits though, considering `git ls-tree` ignores duplicate paths anyway. Likely negligible compared to linter run times, though. What do you think?

sds:
Merged in d88c6c7. Thanks for addressing all those comments!

lib/overcommit/hook/pre_commit/case_conflicts.rb
sds: I think the use of `Set` for `repo_files` is still useful for the constant lookup time to see if an element is present within it. I suppose we could use the ordering of the elements returned by `git ls-tree` but that seems like more trouble than it's worth. The use of `Set#classify` works nicely here.

However, it does look like the set isn't necessary for the `paths` variable. So we can change the first line to:

```ruby
paths = applicable_files.map { |file| File.dirname(file) + File::SEPARATOR }
```

...allowing us to get rid of the `.to_a`.

sds: At the end of the day, not a big deal.

PR #134, opened 2015-03-18
Ensure environment is restored on error

### LINKED ISSUE #129: Unstaged changes lost when pre-commit plugin modified
During a commit I received the message "The following pre-commit plugins have been added, changed, or..." since a plugin had been modified. At that point (before running overcommit --sign pre-commit), I checked git status and noticed unstaged items were lost.

jawshooah:
Fixes #129

There doesn't seem to be a hook_runner_spec or hook_loader_spec anywhere, and I figured that's something you guys would prefer to set up on your own.

### DIFF
--- a/lib/overcommit/hook_runner.rb
+++ b/lib/overcommit/hook_runner.rb
@@ -22,11 +22,13 @@ module Overcommit
       # these calls to allow some sort of "are you sure?" double-interrupt
       # functionality, but until that's deemed necessary let's keep it simple.
       InterruptHandler.isolate_from_interrupts do
-        @context.setup_environment
-        load_hooks
-        result = run_hooks
-        @context.cleanup_environment
-        result
+        begin
+          @context.setup_environment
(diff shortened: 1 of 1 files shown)


lib/overcommit/hook_runner.rb
```diff
@@ -22,11 +22,13 @@ def run
...
-        result = run_hooks
-        @context.cleanup_environment
-        result
+        begin
+          @context.setup_environment
+          load_hooks
+          run_hooks
+        ensure
+          @context.cleanup_environment
+        end
```
sds: This definitely breaks things, because you're no longer returning `result`.

sds:
Yeah, the lack of `hook_runner_spec` is more a result of the fact that it's difficult to test interrupt-handling code, which is a lot of what the `HookRunner` class handles. 

Merged in c3d972d. Thanks for the fix!

PR #141, opened 2015-03-24
Save old hooks on install, restore on uninstall

### LINKED ISSUE #138: Restore old hooks on uninstall
Currently Overcommit clobbers any existing `pre-commit`, `post-checkout`, etc. hooks it finds already in the `.git/hooks` directory on `overcommit --install`. These hooks should be preserved somewhere and restored on `overcommit --uninstall`.

### LINKED ISSUE #139: Config files in subdirectories should inherit from and override the root config
Overcommit should allow `.overcommit.yml` files in subdirectories to override settings in the root config for files in that subdirectory and its children, similar to [RuboCop](https://github.com/bbatsov/rubocop#inheritance).

For example, say I'm using the `Scalastyle` hook, but I'd like to use different `scalastyle-config.xml` files for application and test code. In the root `.overcommit.yml`, I would have

```yaml
Pre-Commit:
  Scalastyle:
    args: ['--config', 'scalastyle-config.xml']
```

And in my `test` directory:

```yaml
inherit_from: ../.overcommit.yml

Pre-Commit:
  Scalastyle:
    args: ['--config', 'test/scalastyle-config.xml']
```

### LINKED ISSUE #140: Hook symlinks remain after `overcommit --uninstall`
The master `overcommit-hook` is removed, but the symlinks remain. Planning to fix this in an upcoming PR that also addresses #138.

jawshooah:
Save existing hooks in '.git/hooks/old-hooks' when running `overcommit --install`, and restore them to `.git/hooks` on `overcommit --uninstall`.

Also remove hook symlinks before the master hook on uninstall.

Closes #138
Closes #140

### DIFF
--- a/lib/overcommit/installer.rb
+++ b/lib/overcommit/installer.rb
@@ -144,7 +144,7 @@ module Overcommit
 
       Dir.chdir(old_hooks_path) do
         Overcommit::Utils.supported_hook_types.each do |hook_type|
-          FileUtils.mv(hook_type, hooks_path)
+          FileUtils.mv(hook_type, hooks_path) if File.exist?(hook_type)
         end
       end
       # Remove old-hooks directory if empty
--- a/lib/overcommit/installer.rb
+++ b/lib/overcommit/installer.rb
@@ -132,6 +132,8 @@ module Overcommit
       Overcommit::Utils.supported_hook_types.each do |hook_type|
         hook_file = File.join(hooks_path, hook_type)
         unless can_replace_file?(hook_file)
+          log.warning "Hook '#{File.expand_path(hook_type)}' already exists and " \
+                      "was not installed by Overcommit. Moving to '#{old_hooks_path}'"
           FileUtils.mv(hook_file, old_hooks_path)
         end
       end
@@ -142,6 +144,8 @@ module Overcommit
     def restore_old_hooks
       return unless File.directory?(old_hooks_path)
 
+      log.log "Restoring old hooks from #{old_hooks_path}"
+
       Dir.chdir(old_hooks_path) do
         Overcommit::Utils.supported_hook_types.each do |hook_type|
           FileUtils.mv(hook_type, hooks_path) if File.exist?(hook_type)
@@ -149,6 +153,8 @@ module Overcommit
       end
       # Remove old-hooks directory if empty
       FileUtils.rmdir(old_hooks_path)
+
+      log.success "Successfully restored old hooks from #{old_hooks_path}"
     end
 
     def uninstall_hook_symlinks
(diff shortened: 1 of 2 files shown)


jawshooah:
Hold off on this one, I forgot to update the specs. I'll take care of it later tonight.

jawshooah:
Specs added, ready for review!

lib/overcommit/installer.rb
```diff
@@ -118,6 +125,32 @@ def can_replace_file?(file)
         overcommit_hook?(file)
     end
 
+    def preserve_old_hooks
+      return unless File.directory?(hooks_path)
+
+      ensure_directory(old_hooks_path)
+      Overcommit::Utils.supported_hook_types.each do |hook_type|
```
sds: Another (perhaps simpler) option is to move `.git/hooks` to `.git/old-hooks`. This way restoring the old hooks is as simple as doing a `rm -rf .git/hooks` and `mv .git/old-hooks .git/hooks`. 

What do you think?

lib/overcommit/installer.rb
```diff
@@ -118,6 +125,32 @@ def can_replace_file?(file)
...
+        hook_file = File.join(hooks_path, hook_type)
+        unless can_replace_file?(hook_file)
+          FileUtils.mv(hook_file, old_hooks_path)
+        end
+      end
+      # Remove old-hooks directory if empty
+      FileUtils.rmdir(old_hooks_path)
```
sds: This looks like it deletes the directory immediately after we populate it?

lib/overcommit/installer.rb
jawshooah: That was my original thought, but I don't think we want to move hooks that aren't supported by Overcommit, and thus wouldn't be overwritten otherwise. So users get to keep their `prepare-commit-msg`, `pre-push`, _etc._ hooks without having to manually replace them from `.git/old-hooks`.

sds: That's fair, although from my perspective it feels like Overcommit should be an "all-or-nothing" buy in. Having said that, I guess it doesn't hurt to keep this behavior.

jawshooah: I think that until (if ever) we have comprehensive support for all hook types, we should be very gentle with existing unsupported hooks. That said, I'm not entirely opposed to moving everything to `.git/old-hooks` or similar as long as we explicitly notify the user of this action.

Speaking of which, there should probably be some logger output regardless of what we decide here. I'll throw that in after a bit.

jawshooah: Ah, I forgot. Another issue with just moving everything to `.git/old-hooks` on install and then doing `rm -rf .git/hooks && mv .git/old-hooks .git/hooks` on uninstall is that the user may have thrown some custom hooks into `.git/hooks` in the meantime, and we don't want to clobber those.

sds:
Merged in e2e198f. Thanks!

PR #149, opened 2015-03-25
Properly handle submodule-only changes

### LINKED ISSUE #148: Unable to commit only submodule updates
In my dotfiles repo, I am trying to update some git submodules, but when I do it, overcommit fails out and I am unable to commit.

```sh
~/dotfiles git:master ❯❯❯ git commit
Running pre-commit hooks

✓ All pre-commit hooks passed

Unable to restore working tree after pre-commit hooks run:
STDOUT:
STDERR:No stash found.
```

At the end of this, I am left with my updated submodules uncommitted and unstaged. This might be happening because overcommit is creating a stash with nothing in it (because submodules are weird), which results in no stash getting created.

I believe I used to be able to do this, so I think this might be a problem with the latest release. overcommit 0.23.0

jawshooah:
Don't clear the working tree or apply the last stash if only submodule changes are staged.

Fixes #148

### DIFF
--- a/spec/overcommit/hook_context/pre_commit_spec.rb
+++ b/spec/overcommit/hook_context/pre_commit_spec.rb
@@ -66,6 +66,30 @@ describe Overcommit::HookContext::PreCommit do
       end
     end
 
+    context 'when only a submodule change is staged' do
+      around do |example|
+        submodule = repo do
+          `git commit --allow-empty -m "Initial commit"`
+        end
+
+        repo do
+          `git submodule add #{submodule} sub &>/dev/null`
+          `git commit -m "Add submodule"`
+          `echo "Hello World" > sub/submodule-file`
+          `git submodule foreach 'git add submodule-file'`
+          `git submodule foreach 'git commit -m "Another commit"'`
+          `git add sub`
+          example.run
+        end
+      end
+
+      it 'keeps staged submodule change' do
+        subject
+        `git diff`.should be_empty
+        File.open('sub/submodule-file', 'r').read.should == "Hello World\n"
+      end
+    end
+
     context 'when a broken symlink is staged' do
       around do |example|
         repo do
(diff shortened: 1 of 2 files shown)


lencioni:
This looks good to me, but @sds has more context and might have better thoughts.

Can you write a test to go along with this?

lencioni:
Oh, also thanks for the speedy pull request!

spec/overcommit/hook_context/pre_commit_spec.rb
```diff
@@ -66,6 +66,30 @@
...
+          `git submodule foreach 'git add submodule-file'`
+          `git submodule foreach 'git commit -m "Another commit"'`
+          `git add sub`
+          example.run
+        end
+      end
+
+      it 'keeps staged submodule change' do
+        subject
+        `git diff`.should be_empty
```
sds: Would it make more sense to check that the submodule change was committed, rather than checking that the diff was empty? It's possible that we simply lost all changes, in which case the diff would still be empty.

jawshooah: But `setup_environment` and `cleanup_environment` run _before_ the commit. Are you suggesting something like this?

```ruby
`git commit -m "Submodule update"`
`git show`.should include('sub')
```

sds: Sorry, poor terminology on my part. We should check that `git diff --cached` has the submodule change in it (i.e. that the changes are still staged). Make sense?

jawshooah: Gotcha, I'll make the change.

spec/overcommit/hook_context/pre_commit_spec.rb
```diff
@@ -66,6 +66,30 @@
...
+          `git submodule foreach 'git commit -m "Another commit"'`
+          `git add sub`
+          example.run
+        end
+      end
+
+      it 'keeps staged submodule change' do
+        expect { subject }.to_not change {
+          (`git diff --cached` =~ /-Subproject commit[\s\S]*\+Subproject commit/).nil?
+        }.from(false)
```
jawshooah: @sds How's this?

sds:
Thanks! Merged in 9172dd36b68b5a761c21c08581b4a66ea7905499.

PR #151, opened 2015-03-26
Fix some existing tests and make others more expressive

### LINKED ISSUE #150: Various spec blocks return raw Boolean values rather than using should-matchers
Throughout the specs, I've noticed a bunch of blocks ending with something like the following:

```ruby
File.open('tracked-file', 'r').read == 'Hello World'
```

This simply returns a Boolean rather than actually performing a test. We know that this doesn't work, because the comparison evaluates to `false` (RHS should be `"Hello World\n"`), but the test still passes. It should be:

```ruby
File.open('tracked-file', 'r').read.should == "Hello World\n"
```

Such invalid tests are peppered throughout the specs. I'm working on a PR now to clean them up.

jawshooah:
The tests in `pre_commit_spec.rb` and `configuration_spec.rb` were actually logically broken, in addition to returning raw Boolean values rather than actually testing anything.

Fixes #150

### DIFF
--- a/spec/overcommit/configuration_spec.rb
+++ b/spec/overcommit/configuration_spec.rb
@@ -196,16 +196,16 @@ describe Overcommit::Configuration do
       context 'and child item contains an array' do
         let(:child) { { 'PreCommit' => { 'SomeHook' => { 'list' => [4, 5] } } } }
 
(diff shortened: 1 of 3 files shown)


sds:
Merged. Thanks for the fixes!

PR #154, opened 2015-04-02
Add support for pre-push hooks

jawshooah:
[Pre-push](https://github.com/git/git/blob/master/templates/hooks--pre-push.sample) hooks are passed the name and url of the remote as arguments, and STDIN contains info about pushed commits, one per line, in the format `local_ref local_sha1 remote_ref remote_sha1`.

### DIFF
--- a/lib/overcommit/hook_context/pre_push.rb
+++ b/lib/overcommit/hook_context/pre_push.rb
@@ -1,0 +1,32 @@
+module Overcommit::HookContext
+  # Contains helpers related to contextual information used by pre-push hooks.
+  class PrePush < Base
+    attr_accessor :args
+
+    def remote_name
+      @args[0]
+    end
+
+    def remote_url
+      @args[1]
+    end
+
+    def pushed_commits
+      input_lines.map do |line|
+        PushedCommit.new(*line.split(' '))
+      end
+    end
+
+    private
+
+    def input_lines
+      @input_lines ||= ARGF.read.split("\n")
+    end
+
+    PushedCommit = Struct.new(:local_ref, :local_sha1, :remote_ref, :remote_sha1) do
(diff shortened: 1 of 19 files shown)


lib/overcommit/hook_context/pre_push.rb
```diff
@@ -0,0 +1,32 @@
...
+    def pushed_commits
+      input_lines.map do |line|
+        PushedCommit.new(*line.split(' '))
+      end
+    end
+
+    private
+
+    def input_lines
+      @input_lines ||= ARGF.read.split("\n")
```
sds: `ARGF` has some additional semantics that aren't necessary here. I'd like it if we just read directly from `STDIN` here.

However, I'd like to take this request a step further by having us inject `STDIN` as a dependency to the hook context itself. Then we could define the `input_lines` helper on `Overcommit::HookContext::Base` and reuse it in `post-rewrite`. To be clear, I'd like to see https://github.com/brigade/overcommit/blob/master/template-dir/hooks/overcommit-hook#L48 become:

```ruby
context = Overcommit::HookContext.create(hook_type, config, ARGV, STDIN)
```

...and have the various `HookContext` classes take an additional argument `input` containing the input stream. I think this will make testing cleaner (it forces us to be explicit about what we're expecting in the input stream, even for hooks that don't read from `STDIN`).

Overcommit will automatically update itself when `overcommit-hook` changes, so this shouldn't have any negative consequences for users when upgrading.

jawshooah: That's how I initially implemented it, but I didn't want this PR to grow beyond its intended scope. But I agree that constructor injection is the way to go here.

sds:
Thanks!

PR #155, opened 2015-04-03
Add pre-push hook to run RSpec test suite

jawshooah:
Add an opt-in hook to run `rspec` before push.

It would be great to enable live terminal output during the hook run to show progress, but I'm not sure how we would go about that with the current framework.

Also, if individual tests write to `stderr`, then simply concatenating `stdout` with `stderr` doesn't preserve the interleaving and isn't an accurate portrayal of the actual output. Again, not sure how to fix this.

### DIFF
--- a/lib/overcommit/hook/pre_push/rspec.rb
+++ b/lib/overcommit/hook/pre_push/rspec.rb
@@ -1,0 +1,12 @@
+module Overcommit::Hook::PrePush
+  # Runs `rspec` test suite before push
+  class Rspec < Base
+    def run
+      result = execute(command)
+      return :pass if result.success?
+
+      output = result.stdout + result.stderr
+      [Overcommit::Hook::Message.new(:error, nil, nil, output)]
+    end
+  end
+end
--- a/lib/overcommit/hook/pre_push/rspec.rb
+++ b/lib/overcommit/hook/pre_push/rspec.rb
@@ -6,7 +6,7 @@ module Overcommit::Hook::PrePush
       return :pass if result.success?
 
       output = result.stdout + result.stderr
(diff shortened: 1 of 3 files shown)


lib/overcommit/hook/pre_push/rspec.rb
```diff
@@ -0,0 +1,12 @@
+module Overcommit::Hook::PrePush
+  # Runs `rspec` test suite before push
+  class Rspec < Base
+    def run
+      result = execute(command)
+      return :pass if result.success?
+
+      output = result.stdout + result.stderr
+      [Overcommit::Hook::Message.new(:error, nil, nil, output)]
```
sds: Since you aren't returning any line information, you can simplify this to:

```ruby
[:fail, output]
```

sds:
Live streaming of terminal output would be quite the undertaking, especially since we want to move towards parallelized hook runs at some point. The complexity it would introduce would also be something I'd rather not introduce if we don't have to.

lib/overcommit/hook/pre_push/rspec.rb
jawshooah: Ah, I thought that form was deprecated.

jawshooah:
True, probably not worth the effort or complexity. And I guess there's really no other way to preserve the order of messages printed to `stdout` and `stderr`.

With that in mind, do you think it would be better to show only one or the other, rather than concatenating the two? E.g. if `stdout` is non-empty, ignore `stderr`.

sds:
I think it depends on the tool. I'm not familiar with when RSpec outputs to `stdout` versus `stderr`. I think we should output whatever information is most informative in helping the developer fix the problem.

Thanks for the feature!

PR #157, opened 2015-04-05
Add ProtectedBranches pre-push hook

jawshooah:
This hook guards against destructive updates to branches specified in the config.

### DIFF
--- a/lib/overcommit/hook/pre_push/protected_branches.rb
+++ b/lib/overcommit/hook/pre_push/protected_branches.rb
@@ -20,7 +20,7 @@ module Overcommit::Hook::PrePush
     def illegal_pushes
       @illegal_pushes ||= pushed_refs.select do |pushed_ref|
         (pushed_ref.deleted? || pushed_ref.forced?) &&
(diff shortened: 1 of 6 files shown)


lib/overcommit/hook/pre_push/protected_branches.rb
```diff
@@ -0,0 +1,27 @@
...
+    private
+
+    def branches
+      @branches ||= config['branches']
+    end
+
+    def illegal_pushes
+      @illegal_pushes ||= pushed_refs.select do |pushed_ref|
+        (pushed_ref.deleted? || pushed_ref.forced?) &&
+          branches.any? { |branch| pushed_ref.remote_ref =~ /#{branch}/ }
```
sds: Are you sure you want to do a regex match here? Suppose I have a branch called `master-and-commander`?

jawshooah: Good catch. Better would probably be `pushed_ref.remote_ref == "refs/heads/#{branch}"`.

sds:
Merged in 256fe2a, thanks!

PR #158, opened 2015-04-06
Add failing test case for restoring unstaged changes

jawshooah:
Add test case for `Overcommit::HookContext::PreCommit#cleanup_environment` to check that unstaged changes are restored when no changes were staged. This test is currently failing.

This is likely a regression introduced by #149. My bad. Fix incoming in a separate PR.

### DIFF
--- a/lib/overcommit/hook_context/pre_commit.rb
+++ b/lib/overcommit/hook_context/pre_commit.rb
@@ -18,9 +18,9 @@ module Overcommit::HookContext
       if !initial_commit? && any_changes?
         @stash_attempted = true
 
+        stash_message = "Overcommit: Stash of repo state before hook run at #{Time.now}"
         result = Overcommit::Utils.execute(
-          %w[git stash save --keep-index --quiet] +
-          ["Overcommit: Stash of repo state before hook run at #{Time.now}"]
+          %w[git stash save --keep-index --quiet] + [stash_message]
         )
 
         unless result.success?
@@ -31,8 +31,7 @@ module Overcommit::HookContext
                 "\nSTDOUT:#{result.stdout}\nSTDERR:#{result.stderr}"
         end
 
-        # False if only submodule references were changed
-        @changes_stashed = modified_files.any?
+        @changes_stashed = `git stash list -1`.include?(stash_message)
       end
 
       # While running the hooks make it appear as if nothing changed
--- a/spec/overcommit/hook_context/pre_commit_spec.rb
+++ b/spec/overcommit/hook_context/pre_commit_spec.rb
@@ -119,11 +119,19 @@ describe Overcommit::HookContext::PreCommit do
       around do |example|
         repo do
           `echo "Hello World" > tracked-file`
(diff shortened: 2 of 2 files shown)


jawshooah:
Pushed the fix on this branch instead of in a separate PR.

sds:
I would rather have the fix for this accompanying the newly introduced tests. However, I can see why you didn't come up with a fix yet, as the problem is rather nuanced.

Some facts:

1. You can `git stash` adding a submodule (you'll get a warning that a non-empty directory&mdash;the submodule directory&mdash;can't be removed, but it works), and restore it with `git stash pop`
2. You can't stash changes to a submodule

I'm a little confused as to the reasoning behind 2, so I've posted a question to the [Git mailing list](https://git-scm.herokuapp.com/community). I would link to the archive but it appears to be down at the moment.

Here is the question I posed to the list:

> I’m having trouble understanding why I cannot stash changes to a submodule.

> When adding a submodule to a repository (`git submodule add
./sub-repo`), I can then run `git stash` and `git stash pop` with
expected results—the submodule disappears and reappears in the working
tree.

> However, when I try stashing an update to a submodule, `git stash`
reports “No local changes to save”. The following shell script
illustrates this behavior:

>```bash
># Create repo
>mkdir test-repo
>cd test-repo
>git init
>git commit --allow-empty -m "Initial commit"

># Create submodule
>mkdir sub-repo
>cd sub-repo
>git init
>git commit --allow-empty -m "Initial commit"
>cd -

># Add submodule
>git submodule add ./sub-repo
>git commit -m "Add submodule"

># Modify submodule
>cd sub-repo
>touch foo
>git add foo
>git commit -m "Submodule changed"
>cd -

># Stash submodule change
>git stash # <---------------------------Displays "No local changes to save”
>```

>I’m trying to wrap my head around why this is the current behavior, as
I suspect this is intentional but it seems unexpected. If anyone can
shed any light on this, I would really appreciate it!

>Thanks,

>Shane

I'll let you know if I hear anything back. However, it looks like we're going to need to special case submodules so that even though git doesn't stash them we "stash" changes on behalf of the user.

sds:
Ah, I see you merged a fix. Great!

However, doesn't this still result in submodule changes being lost, since when we clear the working tree any submodule changes will be destroyed and not restored in the subsequent `git stash pop`?

jawshooah:
The specs for submodule-only changes still pass, so we know that when only submodule changes are staged, everything works as expected. We should add specs for when submodule changes are staged _in addition to_ file-level changes for completeness.

jawshooah:
@sds I've added specs to check that submodule changes are preserved when staged along with file changes. All green.

jawshooah:
You could also post your question to the [git-users](https://groups.google.com/forum/?fromgroups#!forum/git-users) Google group in case the mailing list is down.

sds:
Nice! For some reason I thought I had tested running `git reset --hard` with submodule changes and had confirmed they were lost. Looks like that's not the case (and I've just reconfirmed it).

PR #160, opened 2015-04-06
Fix missing method error in PushedRef#forced?

jawshooah:
I was careless and didn't check whether String had a method `any?`. Thought it was `Enumerable` as in some other languages, but apparently not.

Also added specs for the `PushedRef` class.


sds:
Thanks for your diligence on this! Merged.

Issue #144, opened 2015-03-25
Parallelize hook runs


sds:
When executing hooks, it would be significantly faster to run hooks in parallel rather than serially.

# Benefits
* Faster hook runs that scale with the number of processors on a machine

# Costs
* The `HookRunner` class will need some serious rethinking to support the collection of hook statuses as they are completed in parallel
* This will change overcommit to be a multi-threaded application, so interrupt handling (and signal handlers) will need to handle cases gracefully. Testing this will be difficult.
* If #143 is implemented, parallelism is potentially hindered depending on the dependency graph. This also introduces complexity as we can no longer just run hooks in topological sort order, but would need to trigger dependent hooks to run once all their dependencies passed. 

# Questions

* **What happens if a hook generates files during its run?**
   This is a potentially big issue. Since we're now running hooks in parallel the temporary files generated by some hooks could potentially affect other hook runs. We may have to support a `parallelize` option which can be set to `false` to force serial execution in extreme cases
* **What happens if a hook runs `git` commands?**
  Similar to the file issue above, some `git` commands require exclusive access to the files in the `.git` directory during their run. We'll likely need a `git` helper (similar to `execute`) that serializes requests so hooks don't accidentally trample on each other

Feel free to leave comments and ask questions in this issue.

Bertg:
Yeah. I'm all for this :) :+1:

sds:
This was implemented in #290, and will be released in Overcommit 0.32.0 once it is ready.

PR #162, opened 2015-04-06
Add Spellcheck commit-msg hook

jawshooah:
This hook uses [hunspell](http://hunspell.sourceforge.net/) to check the commit message for potential misspellings, and suggests alternate spellings if available.

### DIFF
--- a/lib/overcommit/hook/commit_msg/spellcheck.rb
+++ b/lib/overcommit/hook/commit_msg/spellcheck.rb
@@ -1,0 +1,37 @@
+require 'tempfile'
+
+module Overcommit::Hook::CommitMsg
+  # Checks the commit message for potential misspellings
+  class Spellcheck < Base
+    Misspelling = Struct.new(:word, :suggestions)
+
+    MISSPELLING_REGEX = /^[&#]\s(?<word>\w+)(?:.+?:\s(?<suggestions>.*))?/
+
+    def run
+      # Remove comments from commit message file
+      update_commit_message(commit_message)
+
+      result = execute(command + [commit_message_file])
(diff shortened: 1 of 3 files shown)


lib/overcommit/hook/commit_msg/spellcheck.rb
```diff
@@ -0,0 +1,37 @@
...
+module Overcommit::Hook::CommitMsg
+  # Checks the commit message for potential misspellings
+  class Spellcheck < Base
+    Misspelling = Struct.new(:word, :suggestions)
+
+    MISSPELLING_REGEX = /^[&#]\s(?<word>\w+)(?:.+?:\s(?<suggestions>.*))?/
+
+    def run
+      # Remove comments from commit message file
+      update_commit_message(commit_message)
```
sds: I'm realizing now that calling a destructive operation on the commit message file as part of a hook run is dangerous if we were to ever parallelize hook runs.

I realize you didn't add this helper, but I also realize that this helper isn't actually used anywhere (until now).

Ideally, we would create a temporary file with the comment lines removed and run the `hunspell` tool against that.

However, I also acknowledge that `commit-msg` hooks would want to be able to modify the commit message at some point.

For now, since this hook doesn't modify the commit message, let's run `hunspell` against a separate temporary file. We still need to think about the `GerritChangeId` hook, which modifies the commit message file directly.

jawshooah: Good call, I've updated the hook accordingly.

As for parallelizing hook runs, any hooks that directly modify the commit message file will probably have to be run serially, which can be specified in the config as discussed in #144.

PR #164, opened 2015-04-06
Add SubmoduleStatus post-checkout, post-commit, post-merge, and post-rewrite hooks

### LINKED ISSUE #145: Add post-checkout hook to remind user to update submodules
When working in a repository with submodules, it would be nice to have a reminder that submodules have been updated.

This can be solved with a `post-checkout` hook that displays a warning if there are submodules that need updating. We'll probably want to have an option that controls whether we also check nested submodules (via the `--recurse` flag on `git submodule status`).

jawshooah:
These hooks warn of any submodules that are uninitialized, out of date with the current index, or contain merge conflicts.

To check submodules recursively, just set `flags: ['--recursive']` in the config.

Closes #145

### DIFF
--- a/lib/overcommit/hook/post_merge/submodule_status.rb
+++ b/lib/overcommit/hook/post_merge/submodule_status.rb
@@ -3,30 +3,7 @@ module Overcommit::Hook::PostMerge
   # notifies the user if any are uninitialized, out of date with
   # the current index, or contain merge conflicts.
   class SubmoduleStatus < Base
-    SUBMODULE_STATUS_REGEX = /
-      ^\s*(?<prefix>[-+U]?)(?<sha1>\w+)
-      \s(?<path>[^\s]+?)
-      (?:\s\((?<describe>.+)\))?$
-    /x
-
-    SubmoduleStatus = Struct.new(:prefix, :sha1, :path, :describe) do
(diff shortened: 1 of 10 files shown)


lib/overcommit/hook/post_merge/submodule_status.rb
```diff
@@ -0,0 +1,55 @@
...
+      [:warn, messages.join("\n")]
+    end
+
+    private
+
+    def parse_submodule_statuses(output)
+      output.scan(SUBMODULE_STATUS_REGEX).map do |prefix, sha1, path, describe|
+        SubmoduleStatus.new(prefix, sha1, path, describe)
+      end
+    end
```
sds: Since we're duplicating this helper `Struct` across each of these hooks, and since submodules are a `git` primitive of sorts, it seems appropriate to extract the regex, struct, and helpers into the `GitRepo` module. That would allow us to DRY up a lot of this. 

What do you think?

jawshooah: Agreed, I've extracted the regex, struct, and helper to `GitRepo`. Also made `recursive` a config setting.

sds:
Thanks!

PR #165, opened 2015-04-07
Add support for pre-rebase hooks

jawshooah:
`pre-rebase` hooks can be used to prevent already-pushed commits from being rebased, to check that rebased commits are not present in other branches, etc.


jawshooah:
Forgot the damned symlink. I'll fix when I get back to my computer.

PR #166, opened 2015-04-07
Extract process logic from `Utils.execute_in_background` into `Subprocess.spawn_detached`

jawshooah:
Move process logic from `Utils.execute_in_background` to `Subprocess.spawn_in_background`. Use `ChildProcess` rather than low-level `Process` module for future-proofing, since `ChildProcess` is Windows-compatible.

Edit: Per the conversation in the comments, this PR now simply extracts the `Process.detach(Process.spawn(...))` logic to `Subprocess.spawn_detached` in the interest of separation of concerns.

### DIFF
--- a/lib/overcommit/utils.rb
+++ b/lib/overcommit/utils.rb
@@ -124,14 +124,14 @@ module Overcommit
       # which we do not need to know the result.
       #
       # @param args [Array<String>]
-      # @return [ChildProcess] thread watching the resulting child process
+      # @return [Thread] thread watching the resulting child process
       def execute_in_background(args)
         if args.include?('|')
           raise Overcommit::Exceptions::InvalidCommandArgs,
                 'Cannot pipe commands with the `execute_in_background` helper'
         end
 
-        Subprocess.spawn_in_background(args)
+        Subprocess.spawn_detached(args)
       end
 
       # Calls a block of code with a modified set of environment variables,
--- a/lib/overcommit/utils.rb
+++ b/lib/overcommit/utils.rb
@@ -124,15 +124,14 @@ module Overcommit
       # which we do not need to know the result.
       #
       # @param args [Array<String>]
-      # @return [Thread] thread watching the resulting child process
+      # @return [ChildProcess] thread watching the resulting child process
       def execute_in_background(args)
         if args.include?('|')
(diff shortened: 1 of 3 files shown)


lib/overcommit/utils.rb
```diff
@@ -124,15 +124,14 @@ def execute(args)
...
+      # @return [ChildProcess] thread watching the resulting child process
       def execute_in_background(args)
         if args.include?('|')
           raise Overcommit::Exceptions::InvalidCommandArgs,
                 'Cannot pipe commands with the `execute_in_background` helper'
         end
 
-        # Dissociate process from parent's input/output streams
-        Process.detach(Process.spawn({}, *args, [:in, :out, :err] => '/dev/null'))
+        Subprocess.spawn_in_background(args)
```
sds: How does this deal with long-running background processes that continue running after the hook run has completed?

The reason we used `Process.detach` here was to ensure that we wouldn't create a [zombie process](http://en.wikipedia.org/wiki/Zombie_process). I'm not sure how `ChildProcess` deals with this situation, but so far it hasn't been necessary to worry about because we don't use it for these sorts of background tasks&mdash;it's always been used in a context that blocks the parent Overcommit process until it completes.

Furthermore, I'm not sure what happens if terminating the parent process terminates the child process. We wouldn't want this to happen for hooks like `IndexTags` which can take many seconds to index large repositories.

jawshooah: We can detach the `ChildProcess` using `process.detach = true` before executing `process.start`. I only omitted that line because `process.exited?` [throws an error](https://github.com/jarib/childprocess/issues/84) when `process` is detached, which caused the spec to fail. Didn't consider the problem of zombie processes.

jawshooah: It doesn't seem like `ChildProcess` [does anything special](https://github.com/jarib/childprocess/blob/master/lib/childprocess/unix/fork_exec_process.rb#L20) to terminate the created process when the parent dies, but that should happen anyway unless the child process has been detached, right?

jawshooah: Regardless, we can just use the original `Process.detach(Process.spawn(...))` logic here and save the `ChildProcess` stuff for later, when we figure out Windows compatibility.

sds:
Merged in 3b67c5d.

PR #167, opened 2015-04-08
Include changes in last commit when running pre-commit hooks after `git commit --amend`

### LINKED ISSUE #146: Detect `git commit --amend` in pre-commit hooks so staged files include everything in commit
Running pre-commit hooks when amending a commit currently behaves a little strangely in that only the changes being added to the commit (not all changes in the commit) are checked by Overcommit.

This sometimes results in weird behavior where depending on the order you add files and amend them to a commit you can get around certain pre-commit hooks. For example, with `Rubocop`'s `MethodLength` cop set to 20 lines, if you add a method that is 19 lines long, commit it, then amend the commit to add another couple of lines to the method body, the `Rubocop` lint message will now become a warning because it is reported on a line that is not currently staged, since it's already in the commit.

# Benefits
* Prevents potential bugs/problems being introduced due to order of changes being amended
* Assuming we can detect when a pre-commit hook is being executed as part of an "amend", this allows us to write a hook that prevents amending commits on certain branches (e.g. `master`) to enforce certain workflows

# Costs
* Unclear if there is a definitive way to detect an amend on pre-commit (if it requires a non-reliable hack, this feature may be dead in the water)
* Significant behavior change, which may be surprising to existing users

Feel free to comment or ask questions in this issue.

jawshooah:
Closes #146

### DIFF
--- a/lib/overcommit/hook_context/pre_commit.rb
+++ b/lib/overcommit/hook_context/pre_commit.rb
@@ -7,7 +7,7 @@ module Overcommit::HookContext
   # This includes staged files, which lines of those files have been modified,
   # etc. It is also responsible for saving/restoring the state of the repo so
   # hooks only inspect staged changes.
-  class PreCommit < Base
+  class PreCommit < Base # rubocop:disable ClassLength
     # Stash unstaged contents of files so hooks don't see changes that aren't
     # about to be committed.
     def setup_environment
--- a/lib/overcommit/hook_context/pre_commit.rb
+++ b/lib/overcommit/hook_context/pre_commit.rb
@@ -10,16 +10,25 @@ module Overcommit::HookContext
   class PreCommit < Base # rubocop:disable ClassLength
     # Returns whether this hook run was triggered by `git commit --amend`
     def amendment?
-      if @amended.nil?
-        cmd = Overcommit::Utils.parent_command
-        @amended = !(/--amend/ =~ cmd).nil?
+      return @amendment unless @amendment.nil?
 
-        amend_alias = `git config --get-regexp '^alias\\.' '--amend'`.
-          slice(/(?<=alias\.)\w+/)
+      cmd = Overcommit::Utils.parent_command
+      amend_pattern = 'commit(\s.*)?\s--amend(\s|$)'
 
-        @amended ||= !(/git #{amend_alias}/ =~ cmd).nil? unless amend_alias.nil?
-      end
-      @amended
+      return @amendment if
+        # True if the command is a commit with the --amend flag
+        @amendment = !(/\s#{amend_pattern}/ =~ cmd).nil?
+
+      # Check for git aliases that call `commit --amend`
+      `git config --get-regexp '^alias\\.' '#{amend_pattern}'`.
+        scan(/alias\.([-\w]+)/). # Extract the alias
+        each do |amend_alias|
+          return @amendment if
+            # True if the command uses a git alias for `commit --amend`
+            @amendment = !(/git\s+#{amend_alias}/ =~ cmd).nil?
+        end
+
+      @amendment
     end
 
     # Stash unstaged contents of files so hooks don't see changes that aren't
--- a/lib/overcommit/hook_context/pre_commit.rb
+++ b/lib/overcommit/hook_context/pre_commit.rb
@@ -82,6 +82,15 @@ module Overcommit::HookContext
         Overcommit::GitRepo.extract_modified_lines(file, staged: true)
     end
 
+    # Returns whether this hook run was triggered by `git commit --amend`
+    def amend?
+      cmd = Overcommit::Utils.last_command
+      amend_alias = `git config --get-regexp '^alias\\.' '--amend'`.
+        slice(/(?<=alias\.)\w+/)
+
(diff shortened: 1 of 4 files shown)


lib/overcommit/hook_context/pre_commit.rb
```diff
@@ -7,7 +7,21 @@ module Overcommit::HookContext
   # This includes staged files, which lines of those files have been modified,
   # etc. It is also responsible for saving/restoring the state of the repo so
   # hooks only inspect staged changes.
-  class PreCommit < Base
+  class PreCommit < Base # rubocop:disable ClassLength
+    # Returns whether this hook run was triggered by `git commit --amend`
+    def amend?
+      if @amended.nil?
+        cmd = Overcommit::Utils.parent_command
+        @amended = !(/--amend/ =~ cmd).nil?
```
sds: While right now there are no other `git` commands that have an `--amend` flag or a flag with `--amend*` as a prefix (that I know of), for the sake of future proofing we should constrain this regex to prevent false positives:

```ruby
/\scommit\s.*--amend(\s|$)/
```

You can also do an early return here since at this point you know this was an amend, saving a shell invocation below. (admittedly a micro-optimization)

```ruby
return @amended if @amended = !(/--amend/ =~ cmd).nil?
```

lib/overcommit/hook_context/pre_commit.rb
```diff
@@ -7,7 +7,21 @@ module Overcommit::HookContext
...
+
+        amend_alias = `git config --get-regexp '^alias\\.' '--amend'`.
+          slice(/(?<=alias\.)\w+/)
```
sds: Aliases can contain hyphens, so you'll need to use more than `\w` here.

lib/overcommit/hook_context/pre_commit.rb
```diff
@@ -7,7 +7,21 @@ module Overcommit::HookContext
...
+
+        @amended ||= !(/git #{amend_alias}/ =~ cmd).nil? unless amend_alias.nil?
```
sds: I don't believe the `unless amend_alias.nil?` here is necessary, since if it's `nil`/`false` the `||` would have resulted in the same resulting outcome. If you implement my early-return suggestion above, you don't even need `||=`.

On another note, I think it make make this code slightly more legible to first-time viewers if we had a quick comment describing what each of these mysterious blocks of regexes are doing.

lib/overcommit/hook_context/pre_commit.rb
```diff
@@ -7,7 +7,21 @@ module Overcommit::HookContext
   # This includes staged files, which lines of those files have been modified,
   # etc. It is also responsible for saving/restoring the state of the repo so
   # hooks only inspect staged changes.
-  class PreCommit < Base
+  class PreCommit < Base # rubocop:disable ClassLength
+    # Returns whether this hook run was triggered by `git commit --amend`
+    def amend?
```
sds: The tense of this verb doesn't seem right given the context we're in. We're trying to return whether this `pre-commit` was the result of an amendment, so it feels like we should call it `amendment?` (making it act like a "is a" predicate). This also reads slightly better when used in `if` statements in hooks.

Naming things is frustrating. :)

spec/overcommit/hook_context/pre_commit_spec.rb
```diff
@@ -7,6 +7,41 @@
...
+          example.run
+        end
+      end
+
+      before do
+        Overcommit::Utils.stub(:parent_command).and_return('git amend')
+      end
+
+      it { should == true }
+    end
```
sds: Should also add a quick test ensuring we don't match silly aliases like:

```ini
[alias]
silly--amend = commit --amend
```

...which _is_ possible if you see my comment above about hypens in aliases.

spec/overcommit/hook_context/pre_commit_spec.rb
```diff
@@ -324,5 +363,76 @@
...
+          example.run
+        end
+      end
+
+      before do
+        context.stub(:amend?).and_return(true)
+      end
+
+      it { should == Set.new(1..3) }
+    end
```
sds: Can we add tests that ensure we're properly merging the two result sets in the event we are amending with additional staged files?

lib/overcommit/hook_context/pre_commit.rb
jawshooah: Pretty sure the `unless` guard is necessary here. If `amend_alias` is `nil`, then it will be interpolated into the regex as an empty string, and we'll get false positives. Could just use `unless amend_alias` instead of `unless amend_alias.nil?` though.

sds: No, you're right. I misread.

spec/overcommit/hook_context/pre_commit_spec.rb
jawshooah: Sure, done in 1377bf2

jawshooah:
@sds Definitely let me know if you think `#amendment?` could be implemented more elegantly. It's kind of a mess at the moment, but with the comments I think it's fairly straightforward.

sds:
Thanks for addressing those comments, @jawshooah. This is great!

Merged in 7f23eb4.

PR #169, opened 2015-04-08
Use ChildProcess to spawn detached processes

jawshooah:
As of [childprocess v0.5.6](https://github.com/jarib/childprocess/commit/14a0d30b0dd5d12808709fa1d82080f12c2e3f99), it is possible to check the exit status of a ChildProcess without an error being thrown.

### DIFF
--- a/lib/overcommit/subprocess.rb
+++ b/lib/overcommit/subprocess.rb
@@ -32,8 +32,12 @@ module Overcommit
       # Spawns a new process in the background using the given array of
       # arguments (the first element is the command).
       def spawn_detached(args)
-        # Dissociate process from parent's input/output streams
-        Process.detach(Process.spawn({}, *args, [:in, :out, :err] => '/dev/null'))
+        process = ChildProcess.build(*args)
+        process.detach = true
+
+        assign_output_streams(process)
+
+        process.start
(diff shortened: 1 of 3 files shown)


lib/overcommit/subprocess.rb
```diff
@@ -32,8 +32,12 @@ def spawn(args)
       # Spawns a new process in the background using the given array of
       # arguments (the first element is the command).
       def spawn_detached(args)
-        # Dissociate process from parent's input/output streams
-        Process.detach(Process.spawn({}, *args, [:in, :out, :err] => '/dev/null'))
+        process = ChildProcess.build(*args)
```
sds: If we are relying on version 0.5.6 or higher of `ChildProcess`, we should update the gemspec accordingly.

jawshooah: Right, I ran `bundle update` but forgot to update the gemspec to match. I'll fix in a bit.

sds: Never mind, I ended up updating it myself in 3c15058 since I was working in the code anyway. Thanks!

sds:
Merged in d9453db. Thanks!

PR #172, opened 2015-04-09
Preserve modification times for both staged and unstaged files

### LINKED ISSUE #159: #setup_environment should store modification times for unstaged files
Currently, `Overcommit::HookContext::PreCommit#setup_environment` only stores the modification times for staged files. If there are unstaged changes, their modification times will change when the stash is reapplied in `#cleanup_environment`.

This is not immediately apparent in the specs, since `#cleanup_environment` is run immediately after `#setup_environment`, and the elapsed time is much less than a second, which is the granularity `File.mtime` provides.

jawshooah:
Closes #159

### DIFF
--- a/lib/overcommit/hook_context/pre_commit.rb
+++ b/lib/overcommit/hook_context/pre_commit.rb
@@ -170,7 +170,10 @@ module Overcommit::HookContext
     def store_modified_times
       @modified_times = {}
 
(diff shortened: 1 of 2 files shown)


sds:
Thanks! Merged in 5d19f58a460f6e03ef020aa791acf6eda0fe15c0.

PR #175, opened 2015-04-13
Add **/*.svg to ImageOptim default include configuration

lencioni:
The image_optim gem can run on SVG files, so we may as well include them
here.


lencioni:
Travis failed, but it looks spurious and unrelated.

PR #176, opened 2015-04-13
Make IndexTags quiet by default

lencioni:
The output of this hook is never important. Making it quiet by default
will help people focus on more important overcommit output.


sds:
Merged in bcbada7e0b632cfb53d9f37bd78f1a68e31fbe3e. Thanks!

PR #182, opened 2015-04-14
Add MergedCommits pre-rebase hook

### LINKED ISSUE #179: Add pre-rebase hook that prevents rebasing already-merged commits
We should add a hook that prevents a user from rebasing commits that have already been merged into one of a specified set of branches (probably just `master` in most cases).

Should be pretty straightforward to implement, we can just check the result of `git branch --contains <commit-sha1>` for each of the rebased commits.

jawshooah:
Closes #179

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -546,3 +546,8 @@ PreRebase:
     requires_files: false
     required: false
     quiet: false
+
+  MergedCommits:
+    enabled: false
+    description: 'Checking for commits that have already been merged'
+    dest_branches: ['master']
--- a/config/default.yml
+++ b/config/default.yml
@@ -550,4 +550,4 @@ PreRebase:
   MergedCommits:
     enabled: false
     description: 'Checking for commits that have already been merged'
-    dest_branches: ['master']
+    branches: ['master']
--- a/lib/overcommit/git_repo.rb
+++ b/lib/overcommit/git_repo.rb
@@ -248,5 +248,16 @@ module Overcommit
       raise Overcommit::Exceptions::GitSubmoduleError,
             "Unable to read submodule information from #{ref}:.gitmodules file: #{ex.message}"
     end
+
+    # Returns the names of all branches containing the given commit.
+    #
+    # @param commit_ref [String] git tree ref that resolves to a commit
+    # @return [Array<String>] list of branches containing the given commit
+    def branches_containing_commit(commit_ref)
+      `git branch --contains #{commit_ref}`.
+        split("\n").
+        reject { |s| s[/detached from/] }.  # ignore detached HEAD
+        map { |s| s[2..-1] }                # trim leading 2 characters
+    end
   end
 end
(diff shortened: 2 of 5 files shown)


jawshooah:
Not sure that `dest_branches` is the most expressive name, but I couldn't come up with anything better on the spot for a list of long-lived branches that are frequent merge targets. Definitely open to suggestions.

lib/overcommit/git_repo.rb
```diff
@@ -248,5 +248,16 @@ def submodules(options = {})
...
+
+    # Returns the names of all branches containing the given commit.
+    #
+    # @param commit_ref [String] git tree ref that resolves to a commit
+    # @return [Array<String>] list of branches containing the given commit
+    def branches_containing_commit(commit_ref)
+      `git branch --contains #{commit_ref}`.
+        split("\n").
+        reject { |s| s[/detached from/] }.  # ignore detached HEAD
+        map { |s| s[2..-1] }                # trim leading 2 characters
```
sds: Sadly, the output of `git branch` can change based on the user's `column.branch` setting, so we need to specify the `--column` flag explicitly to override whatever that is set to.

After playing around with a few of the options the best output I could find was:

```
git branch --contains <hash> --column=dense`
```

...which outputs:

```
  master * current-branch some/other/branch another-branch
```

Since the asterisk marks the current branch and branches can't have spaces, we could `split(/\s+/)` and filter out the asterisk to get all branches.

config/default.yml
```diff
@@ -546,3 +546,8 @@ PreRebase:
     requires_files: false
     required: false
     quiet: false
+
+  MergedCommits:
+    enabled: false
+    description: 'Checking for commits that have already been merged'
+    dest_branches: ['master']
```
sds: I think we're fine to just call this `branches`, given the context and given no other name really jumps out.

lib/overcommit/hook/pre_rebase/merged_commits.rb
```diff
@@ -0,0 +1,29 @@
+module Overcommit::Hook::PreRebase
+  # Prevents rebasing commits that have already been merged into one of
+  # a specified set of destination branches.
+  class MergedCommits < Base
+    def run
+      # Allow rebasing a detached HEAD since no refs are changed.
+      return :pass if detached_head? || illegal_commits.empty?
+
+      message = 'Cannot rebase commits that have already been merged into ' \
+                "one of #{dest_branches.join(',')}"
```
sds: Probably want some whitespace on that join: `join(', ')`

jawshooah: Yep, that was an oversight. Fix incoming.

lib/overcommit/git_repo.rb
jawshooah: Ah, wasn't aware of that option. Still need to snip out the `(detached from <hash>)` entry before splitting on whitespace, though.

sds:
Thanks for addressing those comments. Merged in 7acea7f7584a5fd9a2c1cf9295bd756cda4426a0.

Issue #161, opened 2015-04-06
Windows support?


jawshooah:
Not sure whether this is something you guys care about, as I assume Brigade is an all-Unix shop, but it would be great if this tool supported Windows so that users in corporate environments (read: me) could share it with their coworkers.

Obviously this would require some fairly massive changes, so I understand if this isn't something you want to pursue.

sds:
I think it would be awesome if Overcommit supported Windows. The reality in the past was that since none of the core contributors to Overcommit used Windows, it was hard to justify adding support since no one would be there to support it.

Your contributions over the past few weeks suggest that you care about this tool.

If you are willing to research and explore what's involved in adding Windows support, drive the implementation, and maintain it, I would be willing to merge your pull requests. My hunch is that you would simply need to add some special cases to Overcommit's low-level `git` and process management helpers to get most of the way there. 

Some _known_ unknowns from my perspective (these may turn out to not be an issue, but as I don't use Windows I can't speak to them):

* I'm not sure how Ruby threading support differs for Windows, so if we end up supporting parallelized hook runs this might get tricky (might have to just disable it for Windows)
* I'm not sure how Windows handles process signals, so a lot of the logic we have in place for gracefully handling Ctrl-C (which sadly doesn't have any tests) might need to have an alternate Windows implementation
* I'm not sure if there are any gotchas with `Process.detach` and `Process.spawn` on Windows. Specifically whether `Overcommit::Utils.execute_in_background` will work
* We use the [`childprocess`](https://github.com/jarib/childprocess) gem to manage reading standard out/error streams and obtaining exit codes of processes. I'm not sure if our use of the gem is cross-platform compatible, so there might be some work required in the `Overcommit::Subprocess` helper class

In light of the above, there's probably going to need to be some work to add more tests verifying the existing behavior of Overcommit so we ensure existing functionality isn't broken once we start going down this path.

PR #185, opened 2015-04-17
[WIP] Add support for Windows

jawshooah:
Following up on #161. As the title indicates, this is very much still a work in progress. Figured I'd go ahead and open this to get some feedback while I continue to work on it.

Since Travis does not yet support Windows as a platform, I'm using [Appveyor](https://ci.appveyor.com/project/jawshooah/overcommit) for CI, and a Windows 7 VM for local testing.


jawshooah:
@lencioni Thanks for the comments, keep 'em coming! Keep in mind that this is a wholly unadulterated commit stream, which means I haven't done any rebase/amend cleanup to eliminate obsolete commits, squash related ones together, or elaborate in commit messages.

lencioni:
This is looking really good so far! If you think this might take a while, a lot of these could be merged as you go, which would help prevent bitrot and make future reviews a bit easier. Feel free to submit incremental pull requests of chunks that you feel good about and we'll try to get them merged in quickly.

Thanks so much for working on adding Windows support to overcommit. There are a lot of people who will benefit from your contributions here.

jawshooah:
@lencioni All changes not specifically related to Windows support have been extracted and merged in separate PRs.

jawshooah:
Apparently the `grep` installed with `unxutils` doesn't properly handle the `\s` character class. For example:

```
C:\> echo foo bar>foobar.txt

C:\> grep -IHn "\s" foobar.txt

C:\> grep -IHn " " foobar.txt
foobar.txt:1:foo bar
```

This causes the `TrailingWhitespace`, `MergeConflicts`, and `LocalPathsInGemfile` specs to fail because they use `\s`.

jawshooah:
@sds @lencioni Sorry for the lack of updates on this, I've been busy with other projects and honestly, I'm not sure how to move forward. Integration tests are [failing](https://ci.appveyor.com/project/jawshooah/overcommit/build/178/job/nfg51t6q65634y2r) with output like the following, and I have no idea why:

```
commiting 
  when a hook fails 
'"C:\Ruby193\lib\ruby\gems\1.9.1\bin\ruby.exe"' is not recognized as an internal or external command,
operable program or batch file.
    exits with a non-zero status (FAILED - 1) 
  when no hooks fail
'"C:\Ruby193\lib\ruby\gems\1.9.1\bin\ruby.exe"' is not recognized as an internal or external command,
operable program or batch file.
    exits successfully 
```

The `GerritChangeId` hook simply won't work in a `cmd.exe` environment without directly invoking `sh` or `bash`, which I suppose we could do by changing `command` in the spec when running on Windows.

But I'm at a total loss regarding the `'"C:\Ruby193\lib\ruby\gems\1.9.1\bin\ruby.exe"' is not recognized` failures.

sds:
Thanks for the heads up @jawshooah,

It's interesting that the error message you display has both single and double quotes `'"..."'` around the executable path. I wonder if that has something to do with the problem?

sds:
Added some debug statements to get deeper into this, and it looks like the auto-generated `overcommit.bat` file that RubyGems creates when installing on Windows has a path pointing to a non-existent Ruby:

```
@ECHO OFF
IF NOT "%~f0" == "~f0" GOTO :WinNT
@"C:\Ruby193\lib\ruby\gems\1.9.1\bin\ruby.exe" "C:/Ruby193/lib/ruby/gems/1.9.1/bin/overcommit" %1 %2 %3 %4 %5 %6 %7 %8 %9
GOTO :EOF
:WinNT
@"%~dp0ruby.exe" "%~dpn0" %*
```

This leads us to the RubyGems code that generates that file which looks like it was updated in https://github.com/rubygems/rubygems/commit/dcf1d124 to fix a bug that looks suspiciously like our problem.

I see in our `appveyor.yml` that we've explicitly pinned RubyGems to 2.3.0. That's unfortunate, as it looks like the fix is only available in 2.4.2 or later. What kind of problems where you seeing in later versions?

jawshooah:
There are issues with building native extensions on Windows in newer versions of `gem`. See rubygems/rubygems#977 for a lengthy discussion on the topic. Doesn't look like it's been resolved yet :disappointed:

sds:
Alright, since Ruby 1.9.3 has been [EOLed as of February](https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/), I have no issue with dropping Ruby 1.9.3 from the Windows test suite.

Removing it reveals a number of failures with integration tests I've added recently. I'll try to have a look at those soon and get them fixed.

jawshooah:
Looks like there were some more calls to `File.symlink`/`FileUtils.ln_s` added as well, which should be replaced with `Overcommit::Utils::FileUtils.symlink`.

jawshooah:
@sds Did some cleanup myself, whittled the errors down to the hook signing integration spec and the `execute_in_background` spec.

jawshooah:
Only thing failing now is `hook_signing_spec`. [This build](https://ci.appveyor.com/project/jawshooah/overcommit/build/207/job/dmb2owq29d6s2jtf) includes debugging output so we can get a better sense of what's going wrong.

jawshooah:
Finally, all tests are passing! :beers:

We're not out of the woods yet, though. Take a look at [this test run](https://ci.appveyor.com/project/jawshooah/overcommit/build/218/job/3447kx2rv2vmu43e). You'll notice at the bottom that the `RuboCop` hook failed due to the command being too long. Not only that, but the failure went by unnoticed because we're not checking `result.stderr` in `RuboCop#run`.

This exposes a couple of problems:
* We're not splitting arguments everywhere that we should.
* Many hooks (`RuboCop`, in this case) don't check for error conditions outside of those that are expected when the command executes as it should.

Another problem, unrelated to this particular test run, is that many hooks grab the file name with a regex group that looks something like `^(?<file>[^:]):`. This won't work on Windows, since every path begins with the drive identifier followed by a `:`. We should replace groups defined as above with `^(?<file>.+):`.

sds:
Thanks for your hard work on this, @jawshooah.

I've squashed some of your commits in an effort to quiet a bit of the intermediate work here. Merged in ee50226...627a8c0.

sds:
Is there anything else holding us back from saying Overcommit supports Windows, @jawshooah?

jawshooah:
Given Windows' stricter limitation on command length, we should probably convert all hooks that call external commands to use "splittable" args, and do so for all such hooks in the future.

jawshooah:
Also important to note that the `GIT_TEMPLATE_DIR` method of installation will not work on Windows, since Unix symlinks are not supported.

PR #194, opened 2015-04-25
Include *.svg for XML checks

lencioni:
SVG is an XML-based vector image format. It seems like we should be
running the XML checks on these files.


sds:
Merged in dac3db6.

PR #205, opened 2015-05-02
Update regex to filter out detached HEAD for Git 2.4.0

### LINKED ISSUE #203: Breaking change in detached HEAD reporting in Git 2.4.0
The way Git reports a detached HEAD has [changed](https://github.com/git/git/blob/v2.4.0/Documentation/RelNotes/2.4.0.txt#L28) in 2.4.0. This breaks `GitRepo.branches_containing_commit`, which filters out a detached HEAD using a regex that is no longer accurate.

jawshooah:
Fixes #203


sds:
Merged in 4cb3c10e85a203663eb761592bb35d3c8bcb491c. Thanks!

PR #206, opened 2015-05-02
Set diff.submodule to short for staged submodule test cases

### LINKED ISSUE #204: `diff.submodule` config breaks submodule test cases when set to `log`
The test cases in `spec/overcommit/hook_context/pre_commit_spec.rb` for staged submodule changes break when the `diff.submodule` config setting is set to `log` rather than the default `short`. This is because those test cases rely on `git diff --cached` reporting submodule changes in the `short` format:

```patch
$ git config diff.submodule short
$ git diff --cached
diff --git a/subrepo b/subrepo
index a4955cd..6d692c2 160000
--- a/subrepo
+++ b/subrepo
@@ -1 +1 @@
-Subproject commit a4955cdcd4b6c5bc70a667384feb35340a244929
+Subproject commit 6d692c2e228b0f33104409610353f3e6d3f1a5d2
```

rather than the `log` format:

```patch
$ git config diff.submodule log
$ git diff --cached
Submodule subrepo a4955cd..6d692c2:
  > Another commit
  > Submodule commit
```

jawshooah:
Without this change, these tests will fail if the user has diff.submodule set to 'log' globally.

Fixes #204

### DIFF
--- a/spec/overcommit/hook_context/pre_commit_spec.rb
+++ b/spec/overcommit/hook_context/pre_commit_spec.rb
@@ -191,6 +191,7 @@ describe Overcommit::HookContext::PreCommit do
       end
 
       it 'keeps staged submodule change' do
+        `git config diff.submodule short`
         expect { subject }.to_not change {
           (`git diff --cached` =~ /-Subproject commit[\s\S]*\+Subproject commit/).nil?
         }.from(false)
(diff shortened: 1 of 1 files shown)


sds:
Merged in 8d1d7e13284889b63634a063bd431df21b82c093. Thanks!

Issue #201, opened 2015-05-01
Committing with an empty commit message causes all CommitMsg hooks to blow up


jfelchner:
Just like what it says.  The hooks should probably not fire and just let git handle the empty commit message.

jfelchner:
The various errors:

```
Hook raised unexpected error
undefined method `[]' for nil:NilClass
```

```
Hook raised unexpected error
undefined method `length' for nil:NilClass
```

```
Hook raised unexpected error
undefined method `rstrip' for nil:NilClass
```

sds:
Hey @jfelchner, thanks for the report!

1. We should definitely not crash here, so let's fix that.

2. We should add a hook that reports empty commit messages, as that seems like something you wouldn't want to ever do. I don't think we should not run `CommitMsg` hooks if there is no message; hooks should be written to support that case. Once #143 is implemented we can even have the built-in hooks depend on the hook that checks for empty commit messages so they don't need to check themselves.

jfelchner:
@sds git already handles empty commit messages so no need to handle that in overcommit but whatever you think is best as long as it doesn't crash. :)

sds:
@jfelchner Ah, I didn't actually try this. I see what you mean. Fair enough&mdash;we should just not crash. :)

jawshooah:
Definitely a good idea to add a hook that checks for an empty message, though.

PR #207, opened 2015-05-03
Add EmptyMessage commit-msg hook

jawshooah:
This hook will fail if the commit message is empty or contains only whitespace. Enabled by default, as I suspect this is something most users would never do intentionally.

Refs #201

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -26,6 +26,10 @@ CommitMsg:
     enabled: true
     description: 'Checking subject capitalization'
 
+  EmptyMessage:
+    enabled: true
+    description: 'Checking for empty commit message'
+
   GerritChangeId:
     enabled: false
     description: 'Ensuring Gerrit Change-Id is present'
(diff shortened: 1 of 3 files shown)


lencioni:
:+1:

sds:
Not entirely sure this is necessary, as `git` seems to already prevent you from creating empty commits.

```
$ mkdir test
$ cd test
$ git init
$ git commit --allow-empty -m " "
Aborting commit due to empty commit message.
```

lencioni:
Oh yeah, that's a good point. I wonder if there are any versions in popular use where this isn't the case?

jawshooah:
```
$ git commit --allow-empty --allow-empty-message -m " "
```

lencioni:
Oh, looks like you can pass `--allow-empty-message` in newer versions, so this still probably makes sense.

sds:
Why there exists an `--allow-empty-message` flag in the first place is a bit of a head scratcher.

In any case, merged in 435d9821d151cfb463e4a20097b9d5ad373f30a9.

PR #208, opened 2015-05-05
Add demo GIF to README.md

lencioni:
A little animated GIF that demonstrates Overcommit will help people
understand what the project does and should help its adoption.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@
 `overcommit` is a tool to manage and configure
 [Git hooks](http://git-scm.com/book/en/Customizing-Git-Git-Hooks).
 
+![Demonstration](https://brigade.github.io/overcommit/overcommit.gif)
+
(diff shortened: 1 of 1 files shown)


sds:
Thanks for putting this together, @lencioni. I was thinking something similar, and agree this is an awesome way to quickly show what the tool does in the common case.

My only comment is that it would be great to crop the image so that at the end of the animation your prompt is at the very bottom of the image, rather than have a bunch of empty space. Would make better use of vertical real estate. :)

lencioni:
> My only comment is that it would be great to crop the image so that at the end of the animation your prompt is at the very bottom of the image, rather than have a bunch of empty space. Would make better use of vertical real estate. :)

Yeah, that had crossed my mind. I'll record a new one and force push it to the `gh-pages` branch in a few minutes.

lencioni:
New version is ready now.

PR #212, opened 2015-05-18
Add change log file.

skywinder:
Hi, as I can see, you are carefully fill tags and labels for issues in your repo.

For such cases I create a [github_changelog_generator](https://github.com/skywinder/github-changelog-generator), that generate change log file based on **tags**, **issues** and merged **pull requests** from :octocat: Issue Tracker.

This PR add change log file to your repo (generated by this script).
You can check, how it is look like here: [Change Log](https://github.com/skywinder/overcommit/blob/add-change-log-file/CHANGELOG_AUTO.md)

Some essential features, that has this script:

-  it **exclude** not-related to changelog issues (any issue, that has label `question` `duplicate` `invalid` `wontfix` )
- Distinguish issues **according labels**:
    - Merged pull requests (all `merged` pull-requests)
    - Bug fixes (by label `bug` in issue)
    - Enhancements (by label `enhancement` in issue)
    -   Issues (closed issues `w/o any labels`)
- Generate neat Change Log file according basic [change log guidelines](http://keepachangelog.com).

You can easily update this file in future by simply run script: `github_changelog_generator brigade/overcommit` in your repo folder and it make your Change Log file up-to-date again!

Hope you find this commit as useful. :wink:

P.S.
I know that you already has CHANGELOG.md file but give this script a chance and compare it with yours change log. Hope, you will love it! :blush:

### DIFF
--- a/CHANGELOG_AUTO.md
+++ b/CHANGELOG_AUTO.md
@@ -1,0 +1,657 @@
+# Change Log
+
+## [Unreleased](https://github.com/brigade/overcommit/tree/HEAD)
+
+[Full Changelog](https://github.com/brigade/overcommit/compare/v0.25.0...HEAD)
+
+**Implemented enhancements:**
+
+- Link to hook homepages where available [\#210](https://github.com/brigade/overcommit/issues/210)
+
+**Merged pull requests:**
+
+- Add demo GIF to README.md [\#208](https://github.com/brigade/overcommit/pull/208) ([lencioni](https://github.com/lencioni))
+
+- Add EmptyMessage commit-msg hook [\#207](https://github.com/brigade/overcommit/pull/207) ([jawshooah](https://github.com/jawshooah))
+
+## [v0.25.0](https://github.com/brigade/overcommit/tree/v0.25.0) (2015-05-02)
+
+[Full Changelog](https://github.com/brigade/overcommit/compare/v0.24.0...v0.25.0)
+
+**Implemented enhancements:**
+
+- Define and apply consistent hook naming scheme [\#199](https://github.com/brigade/overcommit/issues/199)
+
+- RuboCop [\#198](https://github.com/brigade/overcommit/issues/198)
+
+- Add pre-commit hook to check Scala formatting with Scalariform [\#192](https://github.com/brigade/overcommit/issues/192)
+
+- Add section that documents the hooks? [\#189](https://github.com/brigade/overcommit/issues/189)
+
+- Add pre-rebase hook that prevents rebasing already-merged commits [\#179](https://github.com/brigade/overcommit/issues/179)
+
+- Add vint pre-commit hook [\#177](https://github.com/brigade/overcommit/issues/177)
+
+**Fixed bugs:**
+
+- `diff.submodule` config breaks submodule test cases when set to `log`  [\#204](https://github.com/brigade/overcommit/issues/204)
+
+- Breaking change in detached HEAD reporting in Git 2.4.0  [\#203](https://github.com/brigade/overcommit/issues/203)
+
+- Committing with an empty commit message causes all CommitMsg hooks to blow up [\#201](https://github.com/brigade/overcommit/issues/201)
+
+- TrailingWhitespace hook fails when un-modified lines have trailing whitespace [\#197](https://github.com/brigade/overcommit/issues/197)
+
+**Closed issues:**
+
+- Pre Push Hook is not running all hooks [\#195](https://github.com/brigade/overcommit/issues/195)
+
+**Merged pull requests:**
+
+- Set diff.submodule to short for staged submodule test cases [\#206](https://github.com/brigade/overcommit/pull/206) ([jawshooah](https://github.com/jawshooah))
+
+- Update regex to filter out detached HEAD for Git 2.4.0 [\#205](https://github.com/brigade/overcommit/pull/205) ([jawshooah](https://github.com/jawshooah))
+
+- Pass commit-msg hooks when commit message is empty [\#202](https://github.com/brigade/overcommit/pull/202) ([jawshooah](https://github.com/jawshooah))
+
+- Move contributing section from readme to CONTRIBUTING.md [\#200](https://github.com/brigade/overcommit/pull/200) ([lencioni](https://github.com/lencioni))
+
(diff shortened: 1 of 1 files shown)


nikolay:
:+1:

sds:
Hey @skywinder, thanks for the pull request. 

`github_changelog_generator` is an interesting project, and I like the goal of making it easier for open source projects to provide high-quality documentation for changes in a project.

I'm less excited about adopting this format for the Overcommit project at this point in time. Some quick reasons:

* There is a lot of "air" (unnecessary blank lines) in the resulting output, which there doesn't appear to be a flag to control. I could write a script that does some post-processing to correct this to my desire, however.
* I would like tighter control over the markup used (e.g. using headings instead of `<strong>` tags via `**..**` Markdown). This is solvable by passing an ERB file (or any template format) describing the desired format, but that is not a feature currently supported. I could also do some post-processing in a separate script.
* Even with the `--exclude-labels` option, there is still a lot of noise for the Overcommit project (lots of issues with incorrect labels). This can be fixed by making a pass at re-labelling everything in the project, but that will take some time and I can't say I'm terribly excited about the idea.
* In order to make the changelog more readable, I will also have to go through and rename issues/pull requests so that they read sensibly when inserted in the changelog (the format of changelog entries is structured so that all [changes start with a verb in the imperative form](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html), which is good practice). I'm also not excited by the thought of this work.

Having said all that, were none of these an obstacle I would be excited to utilize this tool for generating changelog entries automatically. I applaud your efforts to make it easier for open source maintainers to maintain better documentation for the changes in tools. Given that the Overcommit project (in my opinion) has a clear, concise changelog without information overload, I don't see this as something we need to address just yet, but I would like to address it in future.

I would also love to be able to adopt this format going forward, rather than take it on wholesale (as it would avoid a lot of the problems mentioned above). It looks like implementing https://github.com/skywinder/github-changelog-generator/issues/211 would solve this issue, so once that is implemented we can revisit this topic.

skywinder:
@sds Thank you for great description and review. I very appreciate that!

1) Can you clarify, which of blank line is "unnecessary"? I agree with you, it can be more compact.

CHANGELOG_AUTO.md
```diff
@@ -0,0 +1,657 @@
+# Change Log
+
```
skywinder: Like this blank line, right?

CHANGELOG_AUTO.md
```diff
@@ -0,0 +1,657 @@
...
+**Merged pull requests:**
+
+- Add demo GIF to README.md [\#208](https://github.com/brigade/overcommit/pull/208) ([lencioni](https://github.com/lencioni))
+
+- Add EmptyMessage commit-msg hook [\#207](https://github.com/brigade/overcommit/pull/207) ([jawshooah](https://github.com/jawshooah))
+
+## [v0.25.0](https://github.com/brigade/overcommit/tree/v0.25.0) (2015-05-02)
+
+[Full Changelog](https://github.com/brigade/overcommit/compare/v0.24.0...v0.25.0)
```
skywinder: and this

skywinder:
2) about converting all markdown syntax to html tags - you're right. I believe, that should be formatters, that can convert markdown syntax to html. 
It can be implemented as a feature, if there is will be another requests for such option. (however I can't imagine a reason reason convert it in html. Why are you want to do this?)

skywinder:
> Even with the --exclude-labels option, there is still a lot of noise for the Overcommit project. 

Can you provide example of such "noise issues"? And I will try to figure out how to handle it.

> This can be fixed by making a pass at re-labelling everything

No way! This genereator is tend to eliminate any special actions from the user (even in case when issues labels filled correctly (like in yours repo).

And finally, about renaming - yes, it's hardest part, but if you will pay attention to maintain it a little bit (even for important issues, that should appear in change log) - it will not be a problem  anymore.

skywinder:
Ok, I will let you know,when I will implement  skywinder/github-changelog-generator#211. :wink: 
For now- you can try to run `github_changelog_generator --unreleased-only` before place a tag and have ability to get a quick brief, which issues was closed since last version.

skywinder:
And finally, your change log get me one more idea, how to make it better. Look at https://github.com/skywinder/github-changelog-generator/issues/241 it will help you to not break your change log format.

CHANGELOG_AUTO.md
sds: In my case I'm specifically referring to bullet points with blank lines between them. Consider the following two lists:

### List without blank lines
- One
- Two
- Three

### List with blank lines
- One

- Two

- Three

I prefer the former to the latter as it takes up less vertical space, but the tool outputs the latter.

I would say that supporting arbitrary templates would solve this problem (but it is a pretty minor issue).

sds:
Regarding 2) Sorry, I didn't mean supporting HTML output. I mean that right now the different headings are output with a specific markdown format, e.g. 

**Implemented enhancements**

versus:

#### Implemented Enhancements

Again, this is a small point, but it is solvable by allowing arbitrary templates.

Regarding "noise", I'm referring to issues like this: https://github.com/skywinder/overcommit/blame/add-change-log-file/CHANGELOG_AUTO.md#L29

In my mind, this isn't something that needs to be communicated to users of the tool in a change log (it doesn't reflect a change in behavior of the tool itself, but is merely a change in documentation). This can be solved by relabeling that issue with a label like `chore` or something else, so it doesn't show up under enhancements.

skywinder:
@sds 
1. according spaces - yep, they doesn't make sense in markdown output, but it's quite easy to read raw file with this spaces. I will make an option to disable it.

2. Yes, I also thinking about such improvement. I added issue for that ability  skywinder/github-changelog-generator#242

3. regarding "noise" issues - the more suitable solution for such issues - is  just add tag `wontfix` (or smth like `chore`) for example. You're right.

BDW:
You can put more, than 1 tag per issue. And excluded labels have more priority than included. 
So if generator find issue with one of "included" and "excluded" labels together - it exclude this issue.

skywinder:
Ok, since 1.6.1 you can set custom labels for sections:

For your chane log format the script should be running with this options:

`github_changelog_generator --bugs-label "Bug Fixes" --enhancement-label "New Features" --issues-label "Changes"`

PR #215, opened 2015-05-23
Fix parsing of reek output for newer versions of reek

piotrb:
Looks like possibly on newer versions of reek (or it never worked to begin with), the output has an indent the original regex didn't account for. Now it does, also added some additional specs to cover this filename parsing.


sds:
Thanks for the fix, @piotrb!

Merged in 994fbca9dda8c1cb036d6889ec5f9eb253bda525.

piotrb:
Any chance of a release that includes this fix?

PR #217, opened 2015-05-23
Add delegator for PreRebase#detached_head?

jawshooah:
Fixes #216

Issue #216, opened 2015-05-23
Error on pre-rebase hook


jfelchner:
Just updated to the latest overcommit and when I try to rebase I get this:

```
Running pre-rebase hooks
Checking for commits that have already been merged....[MergedCommits] FAILED
Hook raised unexpected error
undefined method `detached_head?' for #<Overcommit::Hook::PreRebase::MergedCommits:0x007fb68137a340>
/usr/local/var/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/overcommit-0.25.0/lib/overcommit/hook/pre_rebase/merged_commits.rb:7:in `run'
/usr/local/var/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/overcommit-0.25.0/lib/overcommit/hook/base.rb:45:in `run_and_transform'
/usr/local/var/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/overcommit-0.25.0/lib/overcommit/hook_runner.rb:107:in `block in run_hook'
/usr/local/var/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/overcommit-0.25.0/lib/overcommit/interrupt_handler.rb:63:in `disable_until_finished_or_interrupted'
/usr/local/var/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/overcommit-0.25.0/lib/overcommit/hook_runner.rb:106:in `run_hook'
/usr/local/var/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/overcommit-0.25.0/lib/overcommit/hook_runner.rb:59:in `block in run_hooks'
/usr/local/var/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/overcommit-0.25.0/lib/overcommit/hook_runner.rb:58:in `each'
/usr/local/var/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/overcommit-0.25.0/lib/overcommit/hook_runner.rb:58:in `run_hooks'
/usr/local/var/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/overcommit-0.25.0/lib/overcommit/hook_runner.rb:39:in `block in run'
/usr/local/var/rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/overcommit-0.25.0/lib/overcommit/interrupt_handler.rb:84:in `isolate_from_interrupts'
[...]

PR #218, opened 2015-05-23
Stub methods of hook context rather than the hook itself

jawshooah:
This will prevent bugs like #216 from cropping up again.

The specs will fail until #217 is merged.


sds:
Merged in c30ff14188ac70f8f9b1dba92e12a1f80f34ec2d. Thanks!

PR #223, opened 2015-06-01
CaseConflicts: Only check against existing files if not initial commit

jawshooah:
Otherwise you get `fatal: Not a valid object name HEAD` since `git ls-tree` requires an existing rev.

This PR also exposes `HookContext::PreCommit#initial_commit?` as a public method with a corresponding delegator to make it accessible to hooks.

### DIFF
--- a/lib/overcommit/hook/pre_commit/case_conflicts.rb
+++ b/lib/overcommit/hook/pre_commit/case_conflicts.rb
@@ -3,8 +3,13 @@ module Overcommit::Hook::PreCommit
   # Adapted from https://github.com/pre-commit/pre-commit-hooks
   class CaseConflicts < Base
     def run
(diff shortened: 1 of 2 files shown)


jawshooah:
Looks like `initial_commit?` must also be defined on the `RunAll` hook context for this to work. @sds, is it worth the code duplication to add that, or should we just use `Overcommit::GitRepo.initial_commit?` in pre-commit hooks?

sds:
Given there are very few built-in hooks that use `inital_commit?`, your suggestion of having them call `Overcommit::GitRepo.initial_commit?` seems reasonable here.

jawshooah:
Agreed, pushed the fix. Should be good to go.

sds:
Merged in d898ff5d485869f30f8f5acc2de74ef0a39664c3. Thanks for the fix!

PR #225, opened 2015-06-03
Exclude files that don't exist from applicable_files

### LINKED ISSUE #219: "No such file or directory @ rb_sysopen" while amending to rename a file
**Problem:** Leo was amending a commit. His original commit added a file. His amend renamed that file. Overcommit was trying to run on the original file, which no longer existed (because it had been renamed).

**Solution:**
```
git stash   # stash current changes
git reset HEAD^   # reset commit he was trying to amend
rm <file that was renamed>
git add -p
git commit   # Use original commit message from gerrit, including change-id
git stash pop
git add -p
git add <new name for renamed file>
git commit --amend
```

This seems related to https://github.com/brigade/overcommit/issues/146.

jawshooah:
Fixes #219

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   empty or contain only whitespace
 * Fix handling of paths with spaces in the name
 * Fix `CaseConflicts` pre-commit hook to not fail on initial commit
+* Fix handling of files removed or renamed in a commit amendment
 
 ## 0.25.0
 
--- a/lib/overcommit/hook/base.rb
+++ b/lib/overcommit/hook/base.rb
@@ -164,6 +164,8 @@ module Overcommit::Hook
     private
 
     def applicable_file?(file)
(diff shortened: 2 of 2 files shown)


lencioni:
:+1: want to add a line to the changelog while you are at it?

jawshooah:
Done!

lencioni:
Thanks for the code @jawshooah!

sds:
Thanks for the pull request, @jawshooah.

On deeper thought, I think I merged this too early. This doesn't solve the underlying problem, which is that `HookContext::PreCommit#modified_files` returns non-existent files. We should probably hoist a fix close to the source of the problem to properly fix this. I'm happy to look into that later.

sds:
Closing the loop on this to say I implemented a more thorough fix with a test in b450815670a4b13b0dc94c5a7af07cd212d04fa4.

PR #229, opened 2015-06-22
Add pre-commit hook for HTMLHint

jawshooah:
[HTMLHint](http://htmlhint.com/) is a linter for HTML.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -195,6 +195,13 @@ PreCommit:
     required_executable: 'grep'
     flags: ['-IHn', "\t"]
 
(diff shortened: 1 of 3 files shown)


jawshooah:
Stripping the color codes is necessary unfortunately, as the developer [doesn't seem interested](yaniswang/htmlhint#14) in adding alternative output formatters.

sds:
Thanks! 

I extracted the `strip_color_codes` helper into the `Utils` module in fa2f208, as it will likely be useful with other hooks in future.

PR #234, opened 2015-06-28
Fix default plugin_directory in Configuration

awseward:
Overcommit::Configuration was using `.githooks` as a default for plugin_directory, instead of `.git-hooks`


sds:
Thanks for making things more consistent here, @awseward.

Merged in 43a4f4c29e3f75c7dfd05452617dd6e46724335e.

Issue #245, opened 2015-07-09
Implement modified_files in PostCheckout, PostMerge, and PostRewrite hook contexts


lencioni:
I would like to write a post-checkout, post-merge, and post-rewrite hook that outputs a warning if one specific file has any changes in it. Something along the lines of:

```ruby
module Overcommit::Hook::PostCheckout
  # Checks if the file has changed
  class FileCheck < Base
    def run
      applicable_files.each do |file|
        if file == 'the file I care about'
          return [:warn, 'the file has changed']
        end
      end
                                                
      :pass
    end
  end
end
```

In my case, I specifically want to print out something like "Run `npm install`" whenever the `npm-shrinkwrap.json` file changes (or possibly even just have overcommit run `npm install` itself).

Since these hook contexts don't implement `modified_files`, the `applicable_files` array is blank and I can't run my check. It looks like `git diff-tree -r --name-only --no-commit-id ORIG_HEAD HEAD` might give you the list of files, so I'm hoping that this won't be too difficult to implement.

jawshooah:
Not related to implementing `modified_files`, but couldn't you skip the `if file == 'foo'` check by just specifying that file in your `includes` config setting?

lencioni:
Yes, but I believe that also depends on `modified_files` working.

jawshooah:
Yup, my comment was tangential.

Regarding implementation, we need to define exactly what it would mean for a file to be modified in this context. Files that are new/different from the last checked-out commit?

jawshooah:
Guess I spoke too soon and took the dive anyway :smile: 

`PostCheckout` now has a `modified_files` helper, and `PostMerge` had one already. `PostRewrite` is all that's left, and I think the new `rewritten_commits` helper may be useful there.

jawshooah:
So `PostRewrite` is a little tricky; this is where we need to be more specific about what we want `modified_files` to mean.

In the case of an amendment, should `modified_files` only include files that were changed between the old and new versions of the commit? Or should it include changes from the old commit as well?

Rebase is even more unclear, as any number of commits could be rewritten. Should we include all files modified in each rewrite?

sds:
> In the case of an amendment, should `modified_files` only include files that were changed between the old and new versions of the commit? Or should it include changes from the old commit as well?

[...]

PR #246, opened 2015-07-10
Implement HookContext::PostCheckout#modified_files

jawshooah:
Return the set of files added or modified between the previous HEAD and new HEAD.

Refs #245


sds:
Thanks @jawshooah. I also added a delegator for `modified_files` in the `PostCheckout::Base` hook in 8ee8707.

jawshooah:
Thanks! But isn't that delegator already defined in `Hook::Base`?

sds:
Indeed it is. Thanks for catching my silliness.

Removed in 86d4ce71c5ff5ead26a0d5578703dea898c7e689.

PR #247, opened 2015-07-10
Add HookContext::PostRewrite#rewritten_commits

jawshooah:
Returns the list of commits rewritten by the amendment or rebase that triggered the hook run.

From the [githooks documentation](https://www.kernel.org/pub/software/scm/git/docs/githooks.html):

> The hook receives a list of the rewritten commits on stdin, in the format

>  `<old-sha1> SP <new-sha1> [ SP <extra-info> ] LF`

> The `extra-info` is again command-dependent. If it is empty, the preceding `SP` is also omitted. Currently, no commands pass any `extra-info`.

### DIFF
--- a/lib/overcommit/hook_context/post_rewrite.rb
+++ b/lib/overcommit/hook_context/post_rewrite.rb
@@ -14,5 +14,18 @@ module Overcommit::HookContext
     def rebase?
       @args[0] == 'rebase'
     end
+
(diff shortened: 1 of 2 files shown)


sds:
Thanks @jawshooah!

Merged in 5b712c4. I renamed the helper struct's fields in f45bf2c.

PR #249, opened 2015-07-13
Implement PostRewrite#modified_files helper

jawshooah:
Collect files modified in each rewritten commit, filtering out directories and nonexistent files.

Refs #245


sds:
This is great, thanks!

Merged as 758348e and e4f4b0e.

PR #251, opened 2015-07-13
Add NpmInstall post-checkout, post-commit, post-merge, and post-rewrite hooks

jawshooah:
Planning on adding similar hooks for package managers like `bundler` and `bower`. Might be worthwhile to add a base `InstallDependencies` module since the code will be pretty much identical.


sds:
Merged in 3b41109 and 6778db2 (squashed so that the build doesn't break if we were to `git bisect`).

Thanks!

PR #257, opened 2015-07-16
Add BowerInstall post-{commit,checkout,merge,rewrite} hooks

jawshooah:
Essentially identical to #251, but for [bower](http://bower.io/).

PR #258, opened 2015-07-16
Add BundleInstall post-{commit,checkout,merge,rewrite} hooks

jawshooah:
Essentially identical to #251 and #257, but for [bundler](http://bundler.io).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Add `PuppetLint` pre-commit hook that checks Puppet code with
   [puppet-lint](http://puppet-lint.com/)
 * Add `BowerInstall` post-checkout, post-commit, post-merge, and post-rewrite hooks
+* Add `BundleInstall` post-checkout, post-commit, post-merge, and post-rewrite hooks
 
 ## 0.27.0
 
--- a/README.md
+++ b/README.md
@@ -336,6 +336,7 @@ follow [proper formatting guidelines](http://tbaggery.com/2008/04/19/a-note-abou
 any time your `HEAD` changes or a file is explicitly checked out.
 
 * [BowerInstall](lib/overcommit/hook/post_checkout/bower_install.rb)
+* [BundleInstall](lib/overcommit/hook/post_checkout/bundle_install.rb)
 * [IndexTags](lib/overcommit/hook/post_checkout/index_tags.rb)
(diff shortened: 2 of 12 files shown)


jawshooah:
@sds Do you think it would be worth it to define a base `InstallDependencies` module to reduce duplication between `NpmInstall`, `BowerInstall`, `BundleInstall` and any similar hooks we may add in the future?

sds:
Come to think of it, what's stopping us from converting these to be treated like [regular git hooks](https://github.com/brigade/overcommit#adding-existing-git-hooks)? All they do is execute a command, so they don't need access to any of the helpers provided by Overcommit. We could just execute the commands directly.

This would also allow us to get rid of all the specs and implementation files.

sds:
Ah, but we'll be passing arguments, so we'll at least need a wrapper to ignore said arguments.

jawshooah:
We'd also need to lift the requirement that ad-hoc hooks be scripts that are tracked in Git.

sds:
I suppose we could lift the restriction that ad hoc hooks be tracked as long as they _aren't_ relative paths or tracked by Git? Basically, we could change the hook signer to only include the contents of the script in the signature if the script resides in the repository (i.e. can be manipulated by an attacker). 

This gives us the flexibility to define inline scripts in our `.overcommit.yml` with `command: ['bash', '-c', 'echo "$@"']` or similar. Since attackers don't have control of the files loaded by your `PATH`, this seems reasonable, and if they change the command in the configuration the signature will still change. We are opening things up for people to expose themselves to danger (i.e. they could execute a bash script like above but have it execute a script in the repository) but I feel there's only so much hand-holding we can do here.

Thoughts?

jawshooah:
I'm all for that idea. We'd be losing the peace of mind that comes with per-hook specs, but I think it's worth it to cut out unnecessary boilerplate.

Regarding security, I agree that at some point we should cut the umbilical cord for the sake of flexibility. It's entirely possible that a user might be running some malicious script that modifies their `PATH` and exposes a dangerous command, but we can't guard against every conceivable attack vector. I think signing scripts inside the repo is probably sufficient.

sds:
Merged in 5f271a8.

jawshooah:
Of course that wouldn't work for commands like [go-lint](https://github.com/brigade/overcommit/blob/master/lib/overcommit/hook/pre_commit/go_lint.rb) that never return non-zero exit codes.

PR #259, opened 2015-07-17
Add pre-commit hook for sqlint

jawshooah:
[sqlint](https://github.com/purcell/sqlint) is a linter for SQL scripts. We recently accepted a SublimeLinter plugin submission for this tool, and though it's still very young, it seems useful.

\cc @purcell @trezona-lecomte


purcell:
Awesome -- so happy to see `sqlint` getting used.

PR #262, opened 2015-07-18
Remove w3c_validators dependency

jawshooah:
This was only used in the specs for `W3cCss` and `W3cHtml`, and only to raise library-specific exceptions. Use mock exceptions instead and remove the dependency.


sds:
Thanks! I had been meaning to do this. We should definitely try to keep these third-party dependencies out to keep things as lean as possible.

Merged in 1e24620ca6d4cdef642592f5dfee4b657f438bb4.

PR #264, opened 2015-07-19
Split file args for RuboCop hook

jawshooah:
Should probably do this for all hooks, but this will at least allow `overcommit --run` to run accurately in Appveyor builds.


sds:
Merged in 0d79818ec2b50fa213016a209f4a383d3045c10f. Thanks!

PR #265, opened 2015-07-19
Include template hook in RuboCop check

jawshooah:
The hooks in `template-dir` were previously not being included in the `RuboCop` check.


sds:
Thanks for catching. Merged in ee502269f97dd758bb7c80d108f098299e845e8d.

PR #266, opened 2015-07-25
Use csv formatter for coffeelint

jawshooah:
This allows us to remove the dependency on the 'json' library and
simplify the message parsing logic. It also makes it easier to parse
concatenated results of multiple calls when splitting argument lists.


sds:
@jawshooah any idea why Appveyor builds take almost two hours to run?

jawshooah:
@sds I'm not sure, mine seem to be completing in around 30 minutes on average: https://ci.appveyor.com/project/jawshooah/overcommit/history

PR #267, opened 2015-07-27
Convert a number of pre-commit hooks to use splittable args

jawshooah:
Very helpful for Windows users :smile:


sds:
Slow build looks related to http://help.appveyor.com/discussions/problems/2548-slow-build-tends-fail-because-of-test-timeout

PR #269, opened 2015-07-28
Add pre-commit hook for recess

jawshooah:
[Recess](https://twitter.github.io/recess/) is a linter for CSS from Twitter.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -368,6 +368,14 @@ PreCommit:
       - 'db/schema.rb'
       - 'db/structure.sql'
 
+  Recess:
+    enabled: false
+    description: 'Analyzing with recess'
+    required_executable: 'recess'
+    install_command: 'npm install -g recess'
+    flags: ['--format=compact', '--stripColors']
+    include: '**/*.css'
+
   Reek:
     enabled: false
     description: 'Analyzing with Reek'
--- a/lib/overcommit/hook/pre_commit/recess.rb
+++ b/lib/overcommit/hook/pre_commit/recess.rb
@@ -1,0 +1,18 @@
+module Overcommit::Hook::PreCommit
+  # Runs `recess` against any modified CSS files.
+  #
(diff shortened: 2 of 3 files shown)


sds:
Not to be overly critical of what qualifies as a useful hook or not, but isn't the Recess project dead? The last commit to the repo was in 2013 and I'm not aware of any active use of it.

Happy to merge if you actually use this yourself, but I would recommend you hop on the SCSS bandwagon and use `scss-lint`. :)

jawshooah:
I'm not using it myself, no. CSSlint development appears to have stalled over the past few months, so I was looking around for alternatives. Had never heard of Recess before, but judging from the star count, I figured some people still found it useful.

You're right, though; no reason to clutter the repo with hooks no one is using.

Regarding SCSS, I unfortunately haven't been able to convince my coworkers that switching to a preprocessor is worth the effort. Does scss-lint work with vanilla CSS?

sds:
Yes, SCSS is a superset of CSS. This makes it ideal for teams thinking about transitioning from vanilla CSS to a preprocessor.

I mention SCSS half-jokingly since we have another project [`scss-lint`](https://github.com/brigade/scss-lint) which supports a wide array of linters. I've heard interesting things about [PostCSS](https://github.com/postcss/postcss) and while I haven't been able to try it out personally it may be an alternative worth exploring (since it too would work with vanilla CSS out of the box, allowing you to extend the standard syntax/features with plugins as you see fit).

For now, I'm going to close this PR without merging. Feel free to reopen if later on you feel this would be useful.

lencioni:
> I've heard interesting things about PostCSS and while I haven't been able to try it out personally it may be an alternative worth exploring (since it too would work with vanilla CSS out of the box, allowing you to extend the standard syntax/features with plugins as you see fit).

I think scss-lint is the only thing that really keeps me from switching to PostCSS.

Issue #271, opened 2015-07-30
ProtectedBranches not protective


ainth:
Given the following `.overcommit.yml`:
```
PrePush:
  ProtectedBranches:
    enabled: true
    description: 'Check for force push to protected branch'
    branches: ['master', 'release', 'integration']
```
When I am on `master` and am 1 commit behind `origin/master` and I make some changes and run:
`git add -A`
`git commit -m "somechanges"`
`git push -f`
Then I see:
```
➜  push-control-test git:(master) git push -f
Running pre-push hooks
Check for force push to protected branch..........[ProtectedBranches] OK

✓ All pre-push hooks passed

```
And the master branch is ruined :(

jawshooah:
I can't reproduce this. Can you create a dummy repository where you can reproduce it and share it here?

sds:
Closing due to lack of response. Feel free to reopen if you can provide a clear list of steps to reproduce.

Thanks!

PR #272, opened 2015-07-30
ProtectedBranches: log pushed_refs in debug mode

jawshooah:
This will help in debugging #271.

### DIFF
--- a/lib/overcommit/hook/pre_push/protected_branches.rb
+++ b/lib/overcommit/hook/pre_push/protected_branches.rb
@@ -2,6 +2,7 @@ module Overcommit::Hook::PrePush
   # Prevents destructive updates to specified branches.
   class ProtectedBranches < Base
     def run
+      log_pushed_refs
       return :pass unless illegal_pushes.any?
 
       messages = illegal_pushes.map do |pushed_ref|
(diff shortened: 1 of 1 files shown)


lib/overcommit/hook/pre_push/protected_branches.rb
```diff
@@ -13,6 +14,15 @@ def run
 
     private
 
+    def log_pushed_refs
+      log_msg = "\nPushed refs:\n\t#{pushed_refs_str}"
+      Overcommit::Utils.log.debug(log_msg)
```
sds: Printing debug messages breaks the output of the hook with the way we currently output `OK`/`FAIL` at the end of the hook's run, so I'd be hesitant to commit this in even if it only appeared with `OVERCOMMIT_DEBUG=1`.

This has come up a few times, so it seems like we need to expose a call that allows a hook to aggregate log messages which are printed in addition to its regular output when debugging is enabled.

jawshooah: Makes sense. We'd need to gate the run in a `begin...ensure` block so that log messages are printed even if the hook run throws an exception.

sds:
Going to close this since the original issue seems quiet, and I'd rather not commit something to master to help what appears to be a user hitting an edge case of some sort, given @jawshooah's [difficulty in reproducing the issue](https://github.com/brigade/overcommit/issues/271#issuecomment-144614754).

PR #277, opened 2015-09-17
Use svg instead of png for Code Climate badge

jawshooah:
`svg` provides a higher-quality badge that's visually consistent with the others.


jawshooah:
Huh, actually it seems that just changing `.png` to `.svg` would do the trick. Fix incoming.

PR #280, opened 2015-09-30
Mention `OVERCOMMIT_DEBUG=1` in CONTRIBUTING.md

lencioni:
In 0.27.0 we added support for the `OVERCOMMIT_DEBUG` environment
variable which toggles the display of verbose output from executed
commands, but didn't document it anywhere. This setting might help
people debug their problems and submit more informative bug reports, so
documenting it here seems to make sense.


sds:
Thanks! Merged in 98c712c77900e43b639eb0fca4d521fef4bae874.

PR #281, opened 2015-10-01
Add NginxTest pre-commit hook to test nginx configs

### LINKED ISSUE #255: Add pre-commit hook to validate Nginx configs
There's an [nginx-lint](https://github.com/temoto/nginx-lint) tool in development, but it's still in its infancy (9 commits in total) and relatively inactive (last commit was in January). Probably better to just use `nginx -t`.

jawshooah:
Finally got around to this. The hook uses `nginx -t` on each config file individually and aggregates any errors.

Closes #255

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -322,6 +322,13 @@ PreCommit:
     required_executable: 'grep'
     flags: ['-IHn', "^<<<<<<<[ \t]"]
 
+  NginxTest:
+    enabled: false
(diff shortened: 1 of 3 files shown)


sds:
Appveyor builds seem to be having trouble, but pretty sure it's unrelated as I've seen it on other changes. Frustrating that they still haven't fixed the issue.

Thanks for the pull request!

PR #285, opened 2015-10-15
Fix tests with git diff.renames option set

aschrab:
If the ~/.gitconfig sets the `diff.renames` option to a true value 4 of
the 5 tests run with `rspec -d renam` fail.  Fix this by adding `R` to
the list of change types included in the `--diff-filter` option.  All
tests continue to pass if this option is left at the default value.


aschrab:
The test failure on AppVeyor looks like a problem with their system unrelated to any changes here.

sds:
Thanks!

PR #286, opened 2015-10-18
Honor `core.commentchar` configuration

aschrab:
Modify commit message parsing to respect the comment character which git has been configured to use.


lib/overcommit/git_config.rb
```diff
@@ -0,0 +1,12 @@
+module Overcommit
+  # Get configuration options from git
+  module GitConfig
+    module_function
+
+    def comment_character(options = '')
```
sds: Typically `options` is the name given to an options hash argument.

Since this option is used only within the context of tests, I don't think it makes sense to have here. Instead, we should write the tests to set the option in `before` blocks (you can create a one-off repo using the `repo` helper so you can modify the repo's `commentchar` option instead of your global one).

sds:
Hey @aschrab,

Thanks for the pull request. Overall this looks good, I just think the `options` argument of the helper is unnecessary, and the tests should set the option explicitly rather than pass it to the helper.

aschrab:
@sds I've updated the code to remove the `options` argument from the helper method as you requested.

PR #290, opened 2015-11-02
Run hooks in parallel

sds:
Add support for running hooks in parallel.

This has been on the [todo](https://github.com/brigade/overcommit/issues/144) list for a while, and while I'm not satisfied with this implementation I think it helps push us forward.

I went with bare threads instead of Celluloid because Celluloid supports only Ruby 2.0+ and I think it was pretty heavyweight for what we're trying to do here.

The biggest challenge here is dealing with interrupts, which is why you see `rescue Interrupt` in a number of places. Since signals aren't yet supported for multi-threaded applications (see the documentation for `Thread.handle_interrupt`), we need to handle a lot of this explicitly.

The output has changed slightly: since we don't run hooks serially, we have to wait until they finish before displaying anything (or else their output is interleaved). I looked into ways of updating the output after the fact (by manipulating the terminal cursor) but it became clear that would be quite the undertaking, so I considered it out of scope.

I need to do some more testing, but putting this up for now so others can have a look and comment.

### DIFF
--- a/lib/overcommit/hook_runner.rb
+++ b/lib/overcommit/hook_runner.rb
@@ -16,6 +16,7 @@ module Overcommit
       @lock = Mutex.new
       @resource = ConditionVariable.new
       @main = ConditionVariable.new
+      @hooks_ready = 0
       @slots_available = @config.concurrency
       @hooks_finished = 0
     end
@@ -58,6 +59,7 @@ module Overcommit
         @printer.start_run
 
         @threads = @hooks.map { |hook| Thread.new(hook, &method(:run_hook)) }
+        wait_for_workers_to_start
 
         begin
           InterruptHandler.disable_until_finished_or_interrupted do
@@ -81,6 +83,14 @@ module Overcommit
       end
     end
 
+    def wait_for_workers_to_start
+      while @hooks_ready < @threads.size
+        @lock.synchronize do
+          @main.wait(@lock)
+        end
+      end
+    end
+
     def start_and_wait_for_workers
       @resource.signal
 
@@ -91,6 +101,8 @@ module Overcommit
 
     def wait_for_slot(hook)
       @lock.synchronize do
+        @main.signal # Tell the main thread that we're ready to run
+        @hooks_ready += 1
         slots_needed = hook.parallelize? ? 1 : @config.concurrency
 
         loop do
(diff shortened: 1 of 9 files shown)


lib/overcommit/hook_runner.rb
```diff
@@ -51,41 +57,95 @@ def run_hooks
...
+    end
+
+    def release_slot(hook)
+      @lock.synchronize do
+        slots_released = hook.parallelize? ? 1 : @config.concurrency
+        @slots_available += slots_released
+        @hooks_finished += 1
+
+        if @hooks_finished < @hooks.size
+          # Signal once. `wait_for_processor` will perform additional signals if
```
jawshooah: Should `wait_for_processor` be `wait_for_slot` here?

sds: Yes, thanks for catching.

lib/overcommit/utils.rb
```diff
@@ -211,6 +211,45 @@ def execute_in_background(args)
...
       end
 
+      # Return the number of processors used by the OS for process scheduling.
+      #
+      # @see https://github.com/grosser/parallel/blob/v1.6.1/lib/parallel/processor_count.rb#L17-L51
+      def processor_count # rubocop:disable all
+        @processor_count ||=
+          begin
+            os_name = RbConfig::CONFIG['target_os']
+            if os_name =~ /mingw|mswin/
```
jawshooah: Could we not use `Overcommit::OS.windows?` here?

sds: Yes, I had simply copied this wholesale from the `parallel` gem (linked above). 

While `os_name` is used further down, the number of systems using Overcommit that would match that `elsif` branch is likely zero for the foreseeable future, so I'll remove it. If it's an issue, someone can make a separate pull request. :)

sds:
Pushed some additional commits along with some documentation. If you get a chance to read the documentation let me know if it makes sense.

Hitting a raise condition in JRuby but I'm pretty sure I know the reason, and I'm glad it was caught in tests because it'll force me to implement things more rigorously.

jawshooah:
The `concurrency` global setting doesn't seem to be explicitly documented outside of `default.yml`, though it is described in the documentation of the `processors` hook setting. Was that intentional?

sds:
Regarding the global `concurrency` option: I've added additional documentation in the README.

The tests are now all passing on Linux. Still waiting on the Appveyor builds but the first few have passed.

Once all those tests have run I'll be merging this and cutting a prerelease gem so people can try it out and report issues.

jawshooah:
Looks like the Appveyor builds passed as well :+1:

lencioni:
\o/

PR #294, opened 2015-11-10
Simplify `BundleCheck` pre-commit hook

deivid-rodriguez:
That part of the hook throws offenses in cases when everything is fine.
Remove it.


sds:
Addressed by ed0fc9c6b2a89f1ea19797741ff3341eefefb978

PR #298, opened 2015-11-22
Add Dogma pre-commit hook to check Elixir source files

dtengeri:
Dogma is a code style linter for Elixir language.
Dogma's source code is available at https://github.com/lpil/dogma
This pre-commit hook executes dogma and reports errors.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -192,6 +192,7 @@ PreCommit:
     enabled: false
     description: 'Analyzing with dogma'
     required_executable: 'mix'
+    flags: ['dogma']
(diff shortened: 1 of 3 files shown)


jawshooah:
Is `mix` sufficient to run the linter? As I gather, it's the equivalent of Ruby's `rake`, so it would depend on the default task. Dogma's own readme suggests running `mix dogma`.

dtengeri:
Yes, you're right, you have to call `mix dogma`. I've included it as a flag in `default.yml`.

PR #304, opened 2015-11-30
Add minitest pre-push hook

danrabinowitz:


config/default.yml
```diff
@@ -783,6 +783,11 @@ PrePush:
     description: 'Running RSpec test suite'
     required_executable: 'rspec'
 
+  Minitest:
+    enabled: false
+    description: 'Running Minitest test suite'
+    command: ['bundle', 'exec', 'rake', 'test']
```
sds: There's nothing about this command that is specific to the `minitest` framework&mdash;it's executing a Rake task that could be executing any code.

danrabinowitz: That makes sense... I was basing this off the rspec command.

Any suggestions on how to run minitest tests (similar to rspec) on PrePush? I'm happy to submit a PR either to the README or the code. I'm just not sure how to run an arbitrary rake task on PrePush without something like this PR.

sds: Under the [Running Your Tests](https://github.com/seattlerb/minitest/blob/9fa5fa541528e5e74bdf4ee0c7692f5e45ace8c5/README.txt#L242) section of the documentation it mentions you can execute the tests directly.

In this case, setting the `required_library` option for the hook to `minitest` and updating the `command` option to `['ruby', '-Ilib:test', 'test']` as mentioned in the documentation.

Basically, invoke the tool directly rather than indirectly, and this pull request is good to go.

danrabinowitz:
Thanks for the guidance. I've updated the PR per your suggestion.

sds:
Thanks! Merged in ed26f3c42d844ff689a8beaa6bc78675506f567f.

config/default.yml
deivid-rodriguez: @sds This doesn't work for me. Here you're not invoking the tool either... you're telling `ruby` to "run a folder"...

```
$ bundle exec ruby -Ilib:test test
/home/deivid/.rubies/ruby-2.3.0/bin/ruby: Is a directory -- test (LoadError)
```

deivid-rodriguez: Something like this works for me:

```
bundle exec ruby -Ilib:test -e 'Dir["test/**/*_test.rb"].each { |test| require File.expand_path(test) }; Minitest.run'
```

deivid-rodriguez: Actually the previous version runs tests twice. This should work instead

```
bundle exec ruby -Ilib:test -rminitest/autorun -e 'Dir["test/**/*_test.rb"].each { |test| require File.expand_path(test) }'
```

danrabinowitz: Maybe submit a PR, so we can discuss there?

deivid-rodriguez: Sure, why not?

PR #307, opened 2015-12-04
Correctly fail hook when rubocop is not installed

fixlr:
When rubocop is not installed, or when the installed version does not match the version expected by Bundler, the 'pre_commit/rubo_cop' hook passes 100% of the time because 'result.stdout' is empty.

This commit updates the hook to fail if anything is present in stderr after the hook runs.

--

**Example:**  I have rubocop v0.29.1 installed locally, but a project's Gemfile.lock depends on v0.34.2. Running `bundle check` correctly identifies the problem:

```
Resolving dependencies......
The following gems are missing
 * rubocop (0.34.2)
Install missing gems with `bundle install`
```

But the overcommit `pre_commit/rubo_cop.rb` hook passes. We catch the rubocop warnings later on CI, but it would be nice to have overcommit surface them locally.

### DIFF
--- a/lib/overcommit/hook/pre_commit/rubo_cop.rb
+++ b/lib/overcommit/hook/pre_commit/rubo_cop.rb
@@ -10,6 +10,7 @@ module Overcommit::Hook::PreCommit
     def run
       result = execute(command, args: applicable_files)
       return :pass if result.success?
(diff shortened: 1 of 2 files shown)


sds:
Thanks for the fix, @fixlr! 

Merged in efa40dc482b75087774e6a53b8428e9a210b4e6a.

PR #309, opened 2015-12-18
Add basic support of rails_best_practices in pre-commit hooks

kovpack:
We were working on one Rails project with @paladiy and found out that `overcommit` had no support of `rails_best_practices` gem, so this is its basic integration.

### DIFF
--- a/lib/overcommit/hook/pre_commit/rails_best_practices.rb
+++ b/lib/overcommit/hook/pre_commit/rails_best_practices.rb
@@ -1,0 +1,32 @@
+module Overcommit
+  module Hook
+    module PreCommit
+      # Runs `rails_best_practices` against Ruby files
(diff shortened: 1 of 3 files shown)


lib/overcommit/hook/pre_commit/rails_best_practices.rb
```diff
@@ -0,0 +1,32 @@
...
+          extract_messages(
+            filter_output(result.stdout),
+            ERROR_REGEXP
+          )
+        end
+
+        private
+
+        def filter_output(stdout)
+          stdout.split("\n").select do |message|
```
jawshooah: Just FYI, this can be done more succinctly with `stdout.split("\n").grep(ERROR_REGEXP)`

PR #310, opened 2015-12-19
Don't count newline against commit subject length

jdewyea:
These changes apply to the `Overcommit::Hooks::CommitMsg::TextWidth` hook.

When specifying a `max_subject_width`, I expect to be able to type out a commit subject as long as the specified max width. However, because the hook counts the newline character, the hook fails.

This pull request takes the behavior used for the commit body -- i.e., not counting a trailing newline against the total width -- and applies that to the commit subject as well.


sds:
Thanks!

Merged in 44fc645 with a fix for Ruby 1.9.3 in ef693fb.

PR #312, opened 2015-12-22
Fix failing tests on Windows

jawshooah:
The `cmd` shell does not treat single-quotes as string delimiters, so we should always use double-quotes in system commands.

Something is wrong with `git-submodule` in the latest release of Git for Windows, so I've added the workaround suggested in git-for-windows/git#181 and in [this post on the mailing list](http://www.spinics.net/lists/git/msg264407.html).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -66,4 +66,5 @@ before_test:
 
 test_script:
   - bundle exec rspec --tty --backtrace --color --format documentation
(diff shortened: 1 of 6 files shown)


sds:
Thanks for the PR, @jawshooah.

Looks like there's still a test failing for Windows (though it was failing before). I'm going to go ahead and merge so we can get some of the benefit.

PR #316, opened 2016-01-02
ProtectedBranches: match glob patterns

### LINKED ISSUE #289: Use regex to match branch names in `PrePush::ProtectedBranches` and `PreRebase::MergedCommits`
Rather than having the user specify exact branch names, which is insufficient in cases where multiple permanent branches exist with similar names (e.g. `release/1.1.x`, `release/1.2.x`, etc), we should allow the user to specify a list of regexes (e.g. `release/.+`).

jawshooah:
This also introduces a new config option `branch_patterns` for `ProtectedBranches`. Its value will be merged with that of `branches` for backwards-compatibility.

Closes #289 (used globs instead of regexes, as that seems a bit more user-friendly).

### DIFF
--- a/lib/overcommit/hook/pre_push/protected_branches.rb
+++ b/lib/overcommit/hook/pre_push/protected_branches.rb
@@ -13,15 +13,22 @@ module Overcommit::Hook::PrePush
 
     private
 
-    def branches
-      @branches ||= config['branches']
-    end
-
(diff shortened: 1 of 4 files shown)


sds:
Merged in e4d71fcda927812d4ceb502b8f9bdcce61e9ec80. Thanks!

PR #317, opened 2016-01-02
PushedRef#forced?: Handle case where git rev-list returns an error

jawshooah:
This may have been the cause of #271.


sds:
Merged in 449c97240730038ac573a039be7badba7c5b7b81.

PR #323, opened 2016-01-18
Add mdl pre-commit hook

jawshooah:
Analyze Markdown files with [mdl/markdownlint][mdl]

[mdl]: https://github.com/mivok/markdownlint

---

Note: The gem is called `mdl`, but the repo is called `markdownlint`. Figured I'd name the hook after the gem, but let me know if I should use `markdownlint` instead.


sds:
Thanks! 

Merged in 687ac61a210d4b16fe26d6e1ac227566a0ef2e33.

PR #332, opened 2016-02-06
Add pre-commit hook ForbiddenBranches

### LINKED ISSUE #327: Add PreCommit hook to disallow committing to specified branches
I work in some projects where I never want to commit to master, but I sometimes mistakenly do this and forget about it, which causes extra friction later when I rebase or create new branches based off of master. It would be nice if I had a hook that prevented me from doing this.

jawshooah:
This hook fails on an attempt to commit to a branch matching any of the glob patterns specified in the `branch_patterns` hook setting.

Closes #327

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -217,6 +217,12 @@ PreCommit:
     description: 'Checking for file execute permissions'
     quiet: true
 
+  ForbiddenBranches:
+    enabled: false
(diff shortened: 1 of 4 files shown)


sds:
Thanks! 

Merged in 556f78981a693116488b25d62465a9b4ccb2c6c4.

PR #334, opened 2016-02-08
update rubocop version to the latest

mhmoudgmal:


sds:
Thanks @mhmoudgmal.

Merged in 677d4c24de8bd1ab9d458bec3e18711b9e5a6d50.

I updated your commit message to match our standards, and highlight the fact that this doesn't affect the version of RuboCop used for hook runs on developer machines, but only the version we use when linting code in Travis.

PR #340, opened 2016-02-11
Add some protips

lencioni:


lib/overcommit/hook/pre_commit/execute_permissions.rb
```diff
@@ -2,6 +2,16 @@ module Overcommit::Hook::PreCommit
   # Checks for files with execute permissions, which are usually not necessary
   # in source code files (and are typically caused by a misconfigured editor
   # assigning incorrect default permissions).
+  #
+  # Protip: if you have some files that you want to allow execute permissions
```
jawshooah: Isn't this just basic functionality available to every hook? Does it warrant a protip?

lib/overcommit/hook/pre_commit/es_lint.rb
```diff
@@ -1,6 +1,16 @@
...
   #
+  # Protip: if you have an npm script set up to run eslint, you can configure
+  # this hook to run eslint via your npm script by using the `command` option in
+  # your .overcommit.yml file. This can be useful if you have some eslint
+  # configuration built into your npm script that you don't want to repeat
+  # somewhere else. Example:
+  #
+  #   EsLint:
+  #     enabled: true
+  #     command: 'npm run lint'
```
jawshooah: Unless I'm mistaken, this should be an array: `['npm', 'run', 'lint']`

lib/overcommit/hook/pre_commit/execute_permissions.rb
lencioni: Yes, it is. But, since people are filing issues asking this question, I think that is evidence enough that it could save people time to have a note here. It helps them apply the feature for this case, which I think is a win.

lib/overcommit/hook/pre_commit/es_lint.rb
lencioni: Yeah, I think you are correct. I'll push an updated version.

sds:
:+1:

PR #341, opened 2016-02-11
Try fixing minitest pre push hook

deivid-rodriguez:
As suggested [here](https://github.com/brigade/overcommit/pull/304#discussion_r52538937), I'm proposing this to fix Minitest's pre-push hook.

Let me know what you think!

### DIFF
--- a/lib/overcommit/hook/base.rb
+++ b/lib/overcommit/hook/base.rb
@@ -17,7 +17,7 @@ module Overcommit::Hook
   class Base # rubocop:disable Metrics/ClassLength
     extend Forwardable
 
-    def_delegators :@context, :modified_files
+    def_delegators :@context, :all_files, :modified_files
     attr_reader :config
 
     # @param config [Overcommit::Configuration]
@@ -155,10 +155,20 @@ module Overcommit::Hook
     # Gets a list of staged files that apply to this hook based on its
     # configured `include` and `exclude` lists.
     def applicable_files
-      @applicable_files ||= modified_files.select { |file| applicable_file?(file) }.sort
+      @applicable_files ||= select_applicable(modified_files)
+    end
+
+    # Gets a list of all files that apply to this hook based on its
+    # configured `include` and `exclude` lists.
+    def included_files
+      @included_files ||= select_applicable(all_files)
     end
 
     private
+
+    def select_applicable(list)
+      list.select { |file| applicable_file?(file) }.sort
+    end
 
     def applicable_file?(file)
       includes = Array(@config['include']).flatten.map do |glob|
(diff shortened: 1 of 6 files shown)


lib/overcommit/hook/base.rb
```diff
@@ -160,22 +160,22 @@ def applicable_files
 
     private
 
-    def applicable_file?(file)
-      includes = Array(@config['include']).flatten.map do |glob|
-        Overcommit::Utils.convert_glob_to_absolute(glob)
-      end
+    def included_files
+      @included_files ||=
+        Overcommit::Utils.expand_relative_glob(@config['include'])
```
sds: We need to `.flatten` the array of includes/excludes since we want to support using YAML references to extend a default list of patterns. See c7471ec37961d734ca7b606be0f35c1a9ca9df51 for context. (it's my fault I didn't add tests for that functionality to catch this)

lib/overcommit/utils.rb
```diff
@@ -270,24 +270,12 @@ def broken_symlink?(file)
...
-      # @param pattern [String]
-      # @param path [String]
-      def matches_path?(pattern, path)
-        File.fnmatch?(pattern, path,
-                      File::FNM_PATHNAME | # Wildcard doesn't match separator
-                      File::FNM_DOTMATCH   # Wildcards match dotfiles
-        )
+      # @return [Array]
+      def expand_relative_glob(globs)
+        Array(globs).map { |glob| Dir[File.join(repo_root, glob)] }.flatten
```
sds: Bear in mind that using `Dir` is dangerous because depending on the glob you'll get _all_ the files that match that glob. So if for example you specify `include: '**/*.js'` and `exclude: 'node_modules/**/*'`, the `include` glob will still match all JS files in `node_modules` slowing things down. This is why you see the implementation used above, where we inspect the glob directly rather than using `Dir`.

sds:
Thanks for taking a stab at fixing this issue, @deivid-rodriguez!

I've left some comments inline, but overall this looks reasonable. The idea of overloading the `command` method is interesting and initially gave me pause, but given it's isolated to one hook I think this is appropriate, as the hook is welcome to implement it's functionality in whatever way makes sense internally to it.

lib/overcommit/utils.rb
sds: I should also mention that the options passed to `File.fnmatch?` are important, as I don't believe `Dir.glob` uses those options by default.

deivid-rodriguez:
Thanks for reviewing @sds! I'll fix your comments.

Regarding extending `command`, I actually have the same feeling. The code came out simpler this way, but this might break a user expectation that the command specified in the config is the actual command being run... Not really sure about it but I can remove the `command` config entirely if we agree it's a better approach.

deivid-rodriguez:
:recycle: I fixed the other issues you mentioned, @sds!

sds:
Thanks!

Rebased and merged in bbbbb3d (nice hash!) and 896de5e.

PR #342, opened 2016-02-16
Print hook information before running it

deivid-rodriguez:
Pre-push hooks that run tests can take a while to execute. Currently
"Running pre-push hooks" is printed and only after running the hook,
"Running Minitest test suite................................[Minitest]
 OK" is printed.

I think it's better to print "Running Minitest test suite..............
..................[Minitest]" before running the hook, and add the
status string afterwards.

### DIFF
--- a/lib/overcommit/hook_runner.rb
+++ b/lib/overcommit/hook_runner.rb
@@ -145,6 +145,8 @@ module Overcommit
     def run_hook(hook) # rubocop:disable Metrics/CyclomaticComplexity
       status, output = nil, nil
 
+      @printer.begin_hook(hook) unless @interrupted
+
       begin
         wait_for_slot(hook)
         return if should_skip?(hook)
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I have the feeling that this won't play well with the multithreaded hook runs... :(

deivid-rodriguez:
In that case, maybe we can change the message from "Running" to "Run" (as in past of "Run") to be more accurate.

sds:
Yes, the parallelization of hook runs makes this concept very difficult. I think the best course of action here (aside from rewriting Overcommit's display module to show the status of multiple hook runs at once–not something I'm particularly keen on), is to reword the hook to say `Ran ...` instead of `Running`.

deivid-rodriguez:
Absolutely, I'll open another PR!

sds:
I've gone ahead and changed the tense to indicative present in 01a6f5a96874c8b28ca61faa2c960a235ba3e797.

PR #344, opened 2016-02-16
Added Regex Pattern based CommitMsg feature

karthik-krishnan:


sds:
Thanks for the pull request, @karthik-krishnan!

Merged in 7fa0e12 with some adjustments in b68ebf1. I renamed the hook to `MessageFormat` (see the commit for details why).

PR #347, opened 2016-02-19
Don't fail on empty message in MessageFormat hook

jawshooah:
The `EmptyMessage` hook already covers this case.


lencioni:
Seems reasonable :+1:

lencioni:
@jawshooah if you haven't already, will you please add an entry to the changelog about this change?

lencioni:
Oh wait, I noticed that the next release will be the first release with this hook, so nevermind.

PR #361, opened 2016-03-29
Silence run status if all hooks are marked quiet

### LINKED ISSUE #357: Add option to hide verbose progress output
Despite setting `quiet: true` for all my hooks, I still see output like this:

```
Running pre-commit hooks

✓ All pre-commit hooks passed

Running commit-msg hooks

✓ All commit-msg hooks passed
```

I only want to see `overcommit` output if errors or warnings are present. This is how other Git hooks work, including `pre-commit` and `fit-commit`.

Some ideas on implementation:
- `log_level` global setting. Users who only want to see exceptional output can change this from `info` to `warn` or `error`.
- `verbose_progress` global setting. Users who only want to see exceptional output will change this from `true` to `false`.
- Output all progress updates to the `debug` log level.
- If `quiet: true` is enabled for a hook, don't output its progress.

m1foley:
Fixes #357

### DIFF
--- a/lib/overcommit/hook_runner.rb
+++ b/lib/overcommit/hook_runner.rb
@@ -53,7 +53,7 @@ module Overcommit
 
     def run_hooks
       if @hooks.any?(&:enabled?)
-        @printer.start_run
+        @printer.start_run if verbose_run_status?
 
         # Sort so hooks requiring fewer processors get queued first. This
         # ensures we make better use of our available processors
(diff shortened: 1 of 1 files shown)


sds:
Thanks for the pull request.

As I suggested in #357, I think a global `quiet` option would be better for this, as it saves you from having to explicitly mark all hooks as `quiet`. I've gone ahead and done that in cefbadc.

m1foley:
Thanks! :+1:

PR #369, opened 2016-04-20
Add GitLfs PrePush hook to replace the stock hook installed by `git lfs install`

alexagranov:
Upon running `overcommit --install`, the previously installed Git Lfs pre_push hook is moved aside to `.git/hooks/old-hooks`.  This hook adds support for GitLfs as a PrePush hook.

Do you think there needs to be more coverage in the specs?  I get nervous when I mock too much...

### DIFF
--- a/lib/overcommit/hook/pre_push/git_lfs.rb
+++ b/lib/overcommit/hook/pre_push/git_lfs.rb
@@ -1,0 +1,43 @@
+module Overcommit::Hook::PrePush
+  # Replaces the pre-push hook installed by `git lfs install` that is then moved aside by
+  # `overcommit --install`
+  #
+  # Enable this hook in .overcommit.yml like so:
+  #
+  #  PrePush:
+  #    GitLfs:
+  #      enabled: true
+  #      command: ['git', 'lfs', 'pre-push']
+  #
+  # @see https://github.com/github/git-lfs
+  class GitLfs < Base
+    # Overcommit expects you to override this method which will be called
+    # everytime your pre-commit hook is run.
+    def run
+      # Create two arrays to hold our error and warning messages.
+      error_lines = []
+      warning_lines = []
+
+      # Check if git-lfs executable exists
+      path_to_git_lfs = `command -v git-lfs`.strip
+      error_lines << "This repository is configured for Git LFS but 'git-lfs' was not found "\
+                     "on your path. If you no longer wish to use Git LFS, disable GitLfs in "\
+                     ".overcommit.yml." unless File.exist?(path_to_git_lfs)
+
+      result = execute(command)
+      error_lines << "'#{command}' returned non-zero exit status: #{result.status}" unless
+        result.success?
+
+      # Overcommit expects 1 of the 3 as return values, `:fail`, `:warn` or `:pass`.
+      # If the hook returns `:fail`, the commit will be aborted with our message
+      # containing the errors.
+      return :fail, error_lines.join("\n") if error_lines.any?
+
+      :pass
+    end
+
+    def command
+      super + ["#{remote_name}", "#{remote_url}"]
+    end
+  end
+end
--- a/lib/overcommit/hook/pre_push/git_lfs.rb
+++ b/lib/overcommit/hook/pre_push/git_lfs.rb
@@ -24,7 +24,7 @@ module Overcommit::Hook::PrePush
                      "on your path. If you no longer wish to use Git LFS, disable GitLfs in "\
                      ".overcommit.yml." unless File.exist?(path_to_git_lfs)
 
-      result = execute(command)
+      result = execute(command, input: pushed_refs)
       error_lines << "'#{command}' returned non-zero exit status: #{result.status}" unless
         result.success?
 
--- a/lib/overcommit/hook/pre_push/git_lfs.rb
+++ b/lib/overcommit/hook/pre_push/git_lfs.rb
@@ -2,42 +2,12 @@ module Overcommit::Hook::PrePush
   # Replaces the pre-push hook installed by `git lfs install` that is then moved aside by
   # `overcommit --install`
   #
-  # Enable this hook in .overcommit.yml like so:
-  #
-  #  PrePush:
-  #    GitLfs:
-  #      enabled: true
-  #      command: ['git', 'lfs', 'pre-push']
-  #
   # @see https://github.com/github/git-lfs
   class GitLfs < Base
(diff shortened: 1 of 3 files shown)


lib/overcommit/hook/pre_push/git_lfs.rb
```diff
@@ -0,0 +1,43 @@
+module Overcommit::Hook::PrePush
+  # Replaces the pre-push hook installed by `git lfs install` that is then moved aside by
+  # `overcommit --install`
+  #
+  # Enable this hook in .overcommit.yml like so:
```
jawshooah: No need for this. You should add the default configuration to `config/default.yml`. There are plenty of examples to pull from there.

lib/overcommit/hook/pre_push/git_lfs.rb
```diff
@@ -0,0 +1,43 @@
...
+  #
+  #  PrePush:
+  #    GitLfs:
+  #      enabled: true
+  #      command: ['git', 'lfs', 'pre-push']
+  #
+  # @see https://github.com/github/git-lfs
+  class GitLfs < Base
+    # Overcommit expects you to override this method which will be called
```
jawshooah: You can remove this comment.

lib/overcommit/hook/pre_push/git_lfs.rb
```diff
@@ -0,0 +1,43 @@
...
+    # everytime your pre-commit hook is run.
+    def run
+      # Create two arrays to hold our error and warning messages.
+      error_lines = []
+      warning_lines = []
+
+      # Check if git-lfs executable exists
+      path_to_git_lfs = `command -v git-lfs`.strip
```
jawshooah: Overcommit already verifies that the command configured as `required_executable` in the hook settings is available in the path, so this check is unnecessary.

lib/overcommit/hook/pre_push/git_lfs.rb
```diff
@@ -0,0 +1,43 @@
...
+      error_lines << "This repository is configured for Git LFS but 'git-lfs' was not found "\
+                     "on your path. If you no longer wish to use Git LFS, disable GitLfs in "\
+                     ".overcommit.yml." unless File.exist?(path_to_git_lfs)
+
+      result = execute(command, input: pushed_refs)
```
jawshooah: `pushed_refs` is an `Array` of `PushedRef` structs. What you want is `input_string`.

lib/overcommit/hook/pre_push/git_lfs.rb
```diff
@@ -0,0 +1,43 @@
...
+      error_lines << "'#{command}' returned non-zero exit status: #{result.status}" unless
```
jawshooah: You should include the actual error message using `result.stdout`, `result.stderr`, or both.

jawshooah: In fact, it's likely enough to just do this:

```rb
return :fail, result.stdout + result.stderr unless result.success?

:pass
```

lib/overcommit/hook/pre_push/git_lfs.rb
alexagranov: Ya this had me scratching my head:
- Git Lfs installation will drop an executable (or symlink) named `/usr/local/bin/git-lfs`
- But their own pre_push hook executes `git lfs pre-push "$@"` which you can see [here](https://github.com/github/git-lfs/blob/8f90f594162d0a7e90e2f5785287f4327f097235/lfs/setup.go#L7)
So I'm left wondering, which should be the `required_executable`, `git` or `git-lfs`??

jawshooah:
With my comments above, the hook could be reduced to the following:

```rb
module Overcommit::Hook::PrePush
  # Replaces the pre-push hook installed by `git lfs install` that is then moved aside by
  # `overcommit --install`
  #
  # @see https://github.com/github/git-lfs
  class GitLfs < Base
    def run
      result = execute(command, args: [remote_name, remote_url], input: input_string)
      return :fail, result.stdout + result.stderr unless result.success?
      :pass
    end
  end
end
```

lib/overcommit/hook/pre_push/git_lfs.rb
jawshooah: `git-lfs`. `git lfs` is just a shim for `git-lfs`, since `git` picks up any external commands available on your path.

alexagranov: Ok, if we go with `required_executable: 'git-lfs'`, I think we'll then need this:

`result = execute(command, args: ['pre-push', remote_name, remote_url], input: input_string)`

as we still need to specify that this is a `pre-push` action.

jawshooah: In the default config, set `flags: ['pre-push']`

jawshooah: And I edited my comment below; it should be `command + [remote_name, remote_url]`. The optional parameter `args` is for a splittable argument list, e.g. a list of files that can be linted independently if the list is too long to fit in a single command.

lib/overcommit/hook/pre_push/git_lfs.rb
alexagranov: Doh!  Looks like its `@context.input_string` though.

lib/overcommit/hook/pre_push/git_lfs.rb
alexagranov: Fixed.

lib/overcommit/hook/pre_push/git_lfs.rb
alexagranov: Fixed.

lib/overcommit/hook/pre_push/git_lfs.rb
alexagranov: Done.

jawshooah:
Just realized I've completely wasted your time. I forgot that [we have something set up for exactly this use case](https://github.com/brigade/overcommit#adding-existing-git-hooks), no PR required. Just drop this into your `.overcommit.yml` and you're good to go:

```yml
PrePush:
  GitLfs:
    enabled: true
    command: ['git-lfs', 'pre-push']
```

alexagranov:
Well that was fun :-)  Might be nice to add another line in that use-case explaining that hooks moved to `old-hooks` can be called exactly this way.

PR #373, opened 2016-04-28
this commit fixes #311

### LINKED ISSUE #311: Overcommit on Windows
I installed the last overcommit version (0.29.1) with Ruby version (2.2.3p173) but the `git commit` doesn't trigger any of pre-commits hooks.
I tested with `overcommit -r` and all the default pre-config hooks were run.

Someone have any ideia what should be the problem?

Thanks in advance!

bsa01:
As I don't think adding the symlinks manually at windows machines, I think using the '/H' option in the mklink command should fix the problem.

### DIFF
--- a/lib/overcommit/utils/file_utils.rb
+++ b/lib/overcommit/utils/file_utils.rb
@@ -53,7 +53,7 @@ module Overcommit::Utils
 
       def win32_mklink_cmd(old_name, new_name)
         Overcommit::Subprocess.spawn(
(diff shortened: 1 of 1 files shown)


bsa01:
Broke the symlink? test.

Meanwhile I found out that on windows you need a special access right to create symlinks (https://superuser.com/questions/124679/how-do-i-create-a-link-in-windows-7-home-premium-as-a-regular-user?answertab=votes#125981) so starting the console with adminstration rights when installing overcommit fixed the missing symlink issue for me.

PR #376, opened 2016-05-01
Add protip about support eslint format

kompiro:
- default format is changed to stylish

### DIFF
--- a/lib/overcommit/hook/pre_commit/es_lint.rb
+++ b/lib/overcommit/hook/pre_commit/es_lint.rb
@@ -10,7 +10,9 @@ module Overcommit::Hook::PreCommit
   #   EsLint:
   #     required_executable: 'npm'
   #     enabled: true
-  #     command: ['npm', 'run', 'lint', '--', '-f', 'compact'] # this hook supports only eslint compact format
+  #     command: ['npm', 'run', 'lint', '--', '-f', 'compact']
+  #
+  # Note: This hook supports only compact format.
   #
   # @see http://eslint.org/
   class EsLint < Base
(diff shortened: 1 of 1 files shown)


lib/overcommit/hook/pre_commit/es_lint.rb
```diff
@@ -8,9 +8,22 @@ module Overcommit::Hook::PreCommit
   # somewhere else. Example:
   #
   #   EsLint:
+  #     required_executable: 'npm'
   #     enabled: true
   #     command: ['npm', 'run', 'lint']
   #
+  # And currently this hook supports only eslint compact format, you can configure
+  # this in package.json. Example:
```
lencioni: You should also be able to configure it for this hook without changing your lint script, like this:

```yaml
command: ['npm', 'run', 'lint', '--', '-f', 'compact']
```

I haven't tested this, but I think it will work.

kompiro: Sorry, `package.json` isn't a lint script. `package.json` is defined about package and development tool dependency. If you'd like to run `npm run lint`, then user has to define it in `package.json`.

If you'd like to read this, [check this.](https://docs.npmjs.com/misc/scripts)

jawshooah: You misunderstand; the change @lencioni proposed should be made in your `.overcommit.yml`, not `package.json`.

kompiro: I like the idea about `.overcommit.yml` to define the args.

But if you'd like to call `eslint` through `npm run`,you should define it in `package.json` like below:

```json
   {
     "scripts": {
       "lint": "eslint"
     },
     "devDependencies": {
       "eslint": "^2.0.0"
     }
   {
```

jawshooah: Yes, that's fine. It also doesn't conflict with the advice above. You could either set `"lint": "eslint -f compact"` in `package.json`, or just use `"lint": "eslint"` and then set `command: ['npm', 'run', 'lint', '--', '-f', 'compact']` in `.overcommit.yml`.

Issue #354, opened 2016-03-09
Allow ProtectedBranches hook to prevent all pushes


jawshooah:
Currently, the `ProtectedBranches` pre-push hook prevents only destructive pushes to specified branches. @shaneog brought up a potential use case in #353 where the hook would prevent *all* pushes to these branches, not just destructive ones.

I propose we add a hook setting `destructive_only` that, when `false`, prohibits all pushes to the configured branches, destructive or not. It would default to `true`, of course.

jawshooah:
Implemented in #378

PR #378, opened 2016-05-09
Add destructive_only flag to PrePush/ProtectedBranches hook

icnagy:
Fixes: #354

Recently came across a situation where it would be beneficial to have a guard against standard git push to protected branches not just the destructive ones. 

The configuration for ProtectedBranch has a new option:

```yaml
PrePush:
  ProtectedBranches:
    enabled: true
    destructive_only: false # <== this is the new option
    branch_patterns:
      - 'master'
      - '*[A-Z]**/*'
```

When set to `destructive_only: false`, ProtectedBranches will check the branch name for all git push operations.
When omitted or set to `destructive_only: true`, ProtectedBranches only kicks in when the operation is destructive (force push, branch deletion).

I am not familiar with RSpec, I hope I got the examples right. Any feedback is welcomed!

### DIFF
--- a/spec/overcommit/hook/pre_push/protected_branches_spec.rb
+++ b/spec/overcommit/hook/pre_push/protected_branches_spec.rb
@@ -41,7 +41,8 @@ describe Overcommit::Hook::PrePush::ProtectedBranches do
     context 'when push is not destructive' do
       context 'and destructive_only set to false' do
         before do
-          subject.stub(:allow_non_destructive?).and_return(true)
+          pushed_ref.stub(:destructive?).and_return(false)
+          subject.stub(destructive_only?: false)
         end
 
         it { should fail_hook }
@@ -49,6 +50,7 @@ describe Overcommit::Hook::PrePush::ProtectedBranches do
 
       context 'and destructive_only set to true' do
         before do
+          subject.stub(destructive_only?: true)
           pushed_ref.stub(:destructive?).and_return(false)
         end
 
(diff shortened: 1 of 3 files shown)


spec/overcommit/hook/pre_push/protected_branches_spec.rb
```diff
@@ -39,19 +39,39 @@
 
   shared_examples_for 'protected branch' do
     context 'when push is not destructive' do
-      before do
-        pushed_ref.stub(:destructive?).and_return(false)
+      context 'and destructive_only set to false' do
+        before do
+          subject.stub(:allow_non_destructive?).and_return(true)
```
sds: Stubbing this helper method is not the greatest, as we're not exercising the logic within. We should be varying the configuration instead.

See https://github.com/brigade/overcommit/blob/583bf0d031d5e/spec/overcommit/hook/pre_commit/forbidden_branches_spec.rb#L6-L13 for an example.

sds:
Thanks for the pull request, @icnagy. Overall looks great, just a comment about some of your stubbing in the tests. (thank you for writing tests!)

It would be great to have @jawshooah have a look as well, but fix that issue and I'll happily merge.

icnagy:
@sds Thank you for the feedback! I could not get my head around the example you gave, instead I went with another approach. Extracted the logical statement that checks for the config option from the new function and stubbed that. Let me know what you think!

Thank you!

sds:
Merged in ab17caf with the spec changes I asked for in 0402044. Hopefully that helps clarify!

PR #380, opened 2016-05-12
Add to run bundle outdated in pre commit

morizyun:
This hook fails on an attempt to commit to keeping latest version gems.

If you have any question or advice, please give me. 🍻


sds:
Thanks for the pull request, @morizyun.

I personally think this would be really noisy as any sizable `Gemfile` (imagine any Rails project) is bound to have at least one gem updated pretty frequently (though I suppose I don't have actual data to back that up).

Still, perhaps others will find this useful. Merged in 83b6033. Thanks!

PR #381, opened 2016-05-16
Make ctags only scan files git knows about

paradox460:
Set ctags to only scan files git knows about. This is useful because it prevents things like vendor or compiled assets from being scanned, speeding up the scan and reducing the size of the tagfile

### DIFF
--- a/libexec/index-tags
+++ b/libexec/index-tags
@@ -8,7 +8,7 @@ set -e
 
 trap "rm -f $GIT_DIR/tags.$$" EXIT
 err_file=$GIT_DIR/ctags.err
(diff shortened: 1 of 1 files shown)


sds:
Hey @paradox460, thanks for the pull request.

I appreciate the direction here, but it's hard to say with 100% certainty whether you want to disable `ctags` scanning of non-git-tracked files by default. It may very well be the case that you want to be able to jump to definitions defined in vendored source code.

The preferred way to avoid this is to add a `.ctags` file in the root of your repo that adds `--exclude` flags as necessary, e.g.

```
--exclude=node_modules
--exclude=public/assets
--exclude=vendor
--exclude=tmp
```

This is more flexible and gives you full control over which files to exclude. 

If there's a strong reason to merge this that I'm still missing, happy to hear your thoughts. Thanks!

Issue #395, opened 2016-06-29
JavaCheckstyle should categorize 'warning' messages correctly


gautamjain:
Checkstyle tags messages (as `INFO`, `WARN`, and `ERROR`) depending on their severity.  However, Overcommit currently treats all messages as an error.  

Will submit a pull request to update this as discussed in
https://github.com/brigade/overcommit/issues/358#issuecomment-204184629

Structon:
Please opt me out of this threat for I am not interested any more

thank you in advance

lencioni:
@Structon it looks like you have chosen to "Watch" this project in the GitHub UI. The best way to prevent future notifications like the one you received for this thread is to remove yourself from the watchers by signing into your GitHub account and changing your watching status from "Watching" to "Not watching".

![screen shot 2016-07-01 at 3 22 24 pm](https://cloud.githubusercontent.com/assets/195534/16535839/af105178-3f9f-11e6-9ba4-9282c4f9d604.png)

I will delete your comment and this one as well to keep this issue clean.

Structon:
OK thank you.

sds:
Implemented in #396. Thanks!

PR #396, opened 2016-06-29
Update JavaCheckstyle to categorize messages depending on type

gautamjain:
Checkstyle generates messages with `ERROR`, `WARN`, or `INFO` tags depending on their severity. Prior to this, all messages (irrespective of their tag) were treated as an error. With this change, messages with `WARN` and `INFO` tags will be handled as Overcommit warnings.

This fixes https://github.com/brigade/overcommit/issues/395

### DIFF
--- a/lib/overcommit/hook/pre_commit/java_checkstyle.rb
+++ b/lib/overcommit/hook/pre_commit/java_checkstyle.rb
@@ -3,18 +3,25 @@ module Overcommit::Hook::PreCommit
   #
   # @see http://checkstyle.sourceforge.net/
   class JavaCheckstyle < Base
-    MESSAGE_REGEX = /^(\[[^\]]+\]\s+)?(?<file>(?:\w:)?[^:]+):(?<line>\d+)/
+    MESSAGE_REGEX = /^(?<type>\[[^\]]+\]\s+)?(?<file>(?:\w:)?[^:]+):(?<line>\d+)/
+
+    MESSAGE_TYPE_CATEGORIZER = lambda do |type|
+      # Type may be nil if checkstyle doesn't output a 'tag'
+      type = type || ""
+
+      type.include?('WARN') || type.include?('INFO') ? :warning : :error
+    end
 
     def run
       result = execute(command, args: applicable_files)
       output = result.stdout.chomp
-      return :pass if result.success?
 
       # example message:
       #   path/to/file.java:3:5: Error message
       extract_messages(
         output.split("\n").grep(MESSAGE_REGEX),
-        MESSAGE_REGEX
+        MESSAGE_REGEX,
+        MESSAGE_TYPE_CATEGORIZER
       )
     end
   end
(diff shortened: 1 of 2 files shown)


lib/overcommit/hook/pre_commit/java_checkstyle.rb
```diff
@@ -3,18 +3,25 @@ module Overcommit::Hook::PreCommit
   #
   # @see http://checkstyle.sourceforge.net/
   class JavaCheckstyle < Base
-    MESSAGE_REGEX = /^(\[[^\]]+\]\s+)?(?<file>(?:\w:)?[^:]+):(?<line>\d+)/
+    MESSAGE_REGEX = /^(?<type>\[[^\]]+\]\s+)?(?<file>(?:\w:)?[^:]+):(?<line>\d+)/
+
+    MESSAGE_TYPE_CATEGORIZER = lambda do |type|
+      # Type may be nil if checkstyle doesn't output a 'tag'
+      type ||= ''
```
sds: Calling `type.to_s` will also convert `nil` to the empty string `''`.

lib/overcommit/hook/pre_commit/java_checkstyle.rb
```diff
@@ -3,18 +3,25 @@ module Overcommit::Hook::PreCommit
...
+
+      type.include?('WARN') || type.include?('INFO') ? :warning : :error
```
sds: This can become:

```ruby
%w[WARN INFO].include?(type.to_s) ? :warning : :error
```

sds: Unless you're trying to handle the case where `type` would also be `WARNING` or `INFORMATION`?

sds: Did you want to still report warnings for `INFO` messages, or just ignore them entirely?

gautamjain: Actually, I wasn't trying to handle `WARNING` or `INFORMATION`.  I'll change it to what you suggested.  Looks _much_ better.

Regarding `INFO` messages, I took a bit of a shortcut and figured we could just report both `INFO` and `WARN` messages as warnings.

gautamjain:
Updated based on your comments.  Had to tweak `MESSAGE_REGEX` because it was capturing tags with the brackets, e.g. `[WARN]` vs `WARN`.

Not sure why the build failed though.  Any idea?

jawshooah:
It's in the log:

> Gem::InstallError: json requires Ruby version ~> 2.0

Looks like the `json` dependency is no longer compatible with Ruby 1.9. Considering that 1.9 has been EOL for a year and a half now, it might be safe to stop supporting it. @sds, what do you think?

sds:
The only reason we hadn't dropped support for 1.9.3 yet was because we hadn't run into an explicit issue with it. Given this issue, I've dropped support for 1.9.3 in 5409330.

sds:
Thanks for your work on this @gautamjain!

Merged in fc1bc2acb01ba63bc4d87c0159cea94263046987.

gautamjain:
Hey @sds, can you cut a new version soon please?  :)

sds:
[0.35.0](https://github.com/brigade/overcommit/releases/tag/v0.35.0) has been released.

gautamjain:
Awesome, thanks!

PR #397, opened 2016-07-09
Add pre-push hook for running test-unit

morizyun:
This hook is to run [test-unit](https://github.com/test-unit/test-unit) like a RSpec or minitest before pushing remote repository.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -838,6 +838,11 @@ PrePush:
     command: ['ruby', '-Ilib:test', '-rminitest', "-e 'exit! Minitest.run'"]
     include: 'test/**/*_test.rb'
 
+  TestUnit:
+    enabled: false
+    description: 'Run test-unit'
+    command: ['rake', 'test']
+
 # Hooks that run during `git rebase`, before any commits are rebased.
 # If a hook fails, the rebase is aborted.
 PreRebase:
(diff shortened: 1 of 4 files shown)


config/default.yml
```diff
@@ -838,6 +838,11 @@ PrePush:
     command: ['ruby', '-Ilib:test', '-rminitest', "-e 'exit! Minitest.run'"]
     include: 'test/**/*_test.rb'
 
+  TestUnit:
+    enabled: false
+    description: 'Run test-unit'
+    command: ['rake', 'test']
```
lencioni: I'm not very familiar with `rake test`, but couldn't it be configured to run anything and not just test-unit? If that is the case, and probably even if it isn't, this hook should be called `RakeTest` instead since that is what it will be running.

lencioni:
A link to this should be added to the readme so that people will know that it exists.

config/default.yml
jawshooah: Alternatively, we could follow the example of `Minitest` and do this:

```yml
TestUnit:
  enabled: false
  description: 'Run Test::Unit test suite'
  command: ['ruby', '-Ilib:test', '-rtest/unit', "-e 'exit! Test::Unit::AutoRunner.run'"]
```

morizyun: Thank you very much for your kind advise. You are right. I fixed `default.yml` which you taught.

morizyun:
Thanks, I added the link to README 😄

sds:
Merged in fb9b2f5ce850b3fc5014ee56ec7585e4c86840cc. Thanks!

PR #398, opened 2016-07-10
Adding pre-commit hook for bundler audit

morizyun:
This hook is to run [bundler-audit](https://github.com/rubysec/bundler-audit) to check vulnerable version of rubyGems.


sds:
Merged in 9855345. Thanks!

Issue #400, opened 2016-07-13
Installing on Windows: symlinking without elevated privileges 


zaps:
Hi, I'm trying to see if I can avoid having Windows users install Overcommit from an elevated command prompt, which is a requirement of the `mklink` command that is used to create symlinks.

In PR #373, @bsa01 modified `lib/overcommit/utils/file_utils.rb` to add the `/H` switch to `mklink`, which creates a hardlink instead of a softlink. This allows a non-admin user to create a link but it broke the `win32_symlink?` method in the file so the PR was closed.

I did an experiment with adding back the `/H` switch and it looks like it did not have an adverse affect on the operation of Overcommit. Granted I did not do extensive edge testing but for typical tasks it seems to work, installs and cleans up after itself. 

Here is an overview of the procedure I performed:

## Current behavior
1. Open a non-elevated Windows command or Git Bash 
2. `git clone http://my-repository`

     contents of `.git/hooks`:
     ```
     applypatch-msg.sample
     commit-msg.sample
     post-update.sample
     pre-applypatch.sample
     pre-commit.sample
     etc
```
3. `overcommit --install`

    contents of `.git/hooks:`
    ```
    applypatch-msg.sample
    commit-msg.sample
    overcommit-hook
    post-update.sample
    pre-applypatch.sample
    pre-commit.sample
    etc
    ```
   And none of the hooks work.

### No custom hooks
This is testing the installer with no custom hooks present in the repo (i.e. with just the default `*.sample` scripts)

1. Delete and reclone the repository
2. Patch the installed overcommit gem (latest 0.34.2) per #373
3. `overcommit --install`

    contents of `.git/hooks:`
    ```
    applypatch-msg.sample
    commit-msg
    commit-msg.sample
    overcommit-hook
    post-update.sample
    post-checkout
    post-commit
    pre-applypatch.sample
    pre-commit
    etc
    ```
   And the hooks work.
4. `overcommit --uninstall`

    contents of `.git/hooks:`
    ```
    applypatch-msg.sample
    commit-msg.sample
    post-update.sample
    pre-applypatch.sample
    pre-commit.sample
    etc
    ```
    The hooks are properly removed by the uninstaller.

### Yes custom hooks
One more time, this time simulating an existing custom `commit-msg` hook

1. Delete and reclone the repository
2. `cd .git\hooks`
3. `move commit-msg.sample commit.msg`
4. `cd ..\..`
5. `overcommit --install`
    
   Output
[...]

PR #403, opened 2016-07-19
Copy hook files instead of linking them

jawshooah:
As explained in #400, creating symlinks on Windows with `mklink` requires admin privileges, which makes Overcommit unusable for non-admin users. Our handling of symlinks on Windows is already brittle enough, so we may as well switch to copying.

We could symlink on *nix and copy on Windows, but that would unnecessarily complicate the code and specs, and I think it's better to keep behavior consistent across platforms anyway.


sds:
Thanks a ton for this, @jawshooah!

PR #404, opened 2016-07-19
appveyor: test against Ruby 2.3

jawshooah:
Also install gem dependencies locally to simplify caching.


sds:
Thanks @jawshooah!

Merged in dfd1fe8 and 78ad621.

PR #406, opened 2016-07-27
Run Brakeman on whole project

janpaul123:
But only when a Ruby file changes.

This is because `--only-files` is dangerous: even if you include every
single Ruby file (as with `overcommit -r`) it doesn't catch
everything. We can either expand it to include all necessary files
(I don't know which they are, and it's brittle as it might change), or
choose the safer option: run Brakeman on the entire project, which for
lots of projects should be pretty fast anyway.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -154,7 +154,7 @@ PreCommit:
     enabled: false
     description: 'Check for security vulnerabilities'
     required_executable: 'brakeman'
-    flags: ['--exit-on-warn', '--quiet', '--summary', '--only-files']
+    flags: ['--exit-on-warn', '--quiet', '--summary']
     install_command: 'gem install brakeman'
     include:
       - '**/*.rb'
--- a/lib/overcommit/hook/pre_commit/brakeman.rb
+++ b/lib/overcommit/hook/pre_commit/brakeman.rb
@@ -1,10 +1,10 @@
 module Overcommit::Hook::PreCommit
-  # Runs `brakeman` against any modified Ruby/Rails files.
+  # Runs `brakeman` whenever Ruby/Rails files change.
   #
   # @see http://brakemanscanner.org/
   class Brakeman < Base
     def run
-      result = execute(command + [applicable_files.join(',')])
+      result = execute(command)
       return :pass if result.success?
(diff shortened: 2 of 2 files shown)


sds:
FWIW, this sets you up to have inconsistent hook runs in the event you are running the hooks in a repository with untracked files (which can affect the outcome of the hook run). You can solve this by removing the `include` option so that any/all files changes trigger it (otherwise users editing views may be confused as to why this hook doesn't run when they edit an ERB/HAML/Slim file).

This hook has never felt like it belongs in the pre-commit cycle, but since it's opt-in I have no hard objections to this.

janpaul123:
Ah, that's a good point (the untracked files), that not good either. Maybe Brakeman is just not suitable to be used with overcommit and should we remove it?

sds:
> Maybe Brakeman is just not suitable to be used with overcommit and should we remove it?

I agree that Brakeman doesn't really fit within the Overcommit model, but since it's disabled by default I'm OK continuing to have it in Overcommit in the event others find some value in it.

sds:
It looks like @jfelchner chose to add the `--only-files` flag in the [original commit](https://github.com/brigade/overcommit/commit/829823df8f7759ad3bc55596d5f89f68b6649e5d), despite being recommended against it by the Brakeman maintainers in https://github.com/presidentbeef/brakeman/issues/467.

I don't have any skin in this game, as I don't think a pre-commit hook is the right place for these kind of in-depth security checks that require the entire repository to properly function (they should instead be a separate job in your CI environment).

Happy to merge if you feel this is the right call, otherwise we can leave things as-is and allow developers to customize the flags in their local `.overcommit.yml` files if they want to scan the entire repository.

jfelchner:
@sds I've since changed my view on this and am 100% ok with removing Brakeman integration or removing the `--only-files` flag or whatever.  I run this manually now.

janpaul123:
Cool, thanks! 👍

zoras:
@sds IMHO Instead of removing `brakeman` hook completely, it should be run as a `PrePush` hook just like with `RSpec` so that all the files of the projects are scanned.

PR #408, opened 2016-07-30
Add commit-msg hook for commitplease

### LINKED ISSUE #405: Add CommitMsg hook for commitplease
The commitplease package validates commit messages. It would be nice to be able to use this via Overcommit.

https://github.com/jzaefferer/commitplease

all3fox:
[Commitplease][1] is an npm module that uses commit-msg hook to check that a commit message follows a certain style guide.

I have described how [it works with overcommit][2] and would be grateful if you could sanity-check that :)

Closes #405
Refs jzaefferer/commitplease#72

[1]: https://github.com/jzaefferer/commitplease
[2]: https://github.com/jzaefferer/commitplease/issues/72#issuecomment-235827024

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -114,6 +114,12 @@ CommitMsg:
     enabled: true
     description: 'Check for trailing periods in subject'
 
+  Commitplease:
+    enabled: false
+    description: 'Analyze with Commitplease'
+    required_executable: './node_modules/.bin/commitplease'
+    install_command: 'npm install --save-dev commitplease'
+
 # Hooks that are run after `git commit` is executed, before the commit message
 # editor is displayed. These hooks are ideal for syntax checkers, linters, and
 # other checks that you want to run before you allow a commit object to be
--- a/lib/overcommit/hook/commit_msg/commitplease.rb
+++ b/lib/overcommit/hook/commit_msg/commitplease.rb
@@ -1,0 +1,12 @@
+module Overcommit::Hook::CommitMsg
+  # Check that a commit message conforms to a certain style
+  class Commitplease < Base
+    def run
+      result = execute(command)
+      output = result.stderr
+      return :pass if result.success? && output.empty?
+
+      [:fail, output]
+    end
+  end
+end


config/default.yml
```diff
@@ -114,6 +114,12 @@ CommitMsg:
     enabled: true
     description: 'Check for trailing periods in subject'
 
+  Commitplease:
+    enabled: false
+    description: 'Analyze with Commitplease'
+    required_executable: './node_modules/.bin/commitplease'
```
sds: I've quickly skimmed the thread https://github.com/jzaefferer/commitplease/issues/72#issuecomment-235827024 and I'm not entirely clear on why you are accessing this via a local `node_modules` directory instead of via your `PATH`. If you look at the `EsLint` hook, we refer to the [`eslint` executable](https://github.com/brigade/overcommit/blob/8edb55305b236908cf74162f0a05dba976ab0609/config/default.yml#L234-L236) instead of its path within the `node_modules` directory.

Has this been tested? Does it work?

sds:
Thanks for the pull request, @all3fox!

I'm a little confused with the decision to use a local `node_modules` directory instead of a globally-installed executable like we do with other hooks (note that developers can always switch their configuration to use the local executable in their `node_modules` directory if they so choose).

all3fox:
Updated this PR based on[ @lencioni and @jawshooah help][1]

> I'm a little confused with the decision to use a local node_modules directory instead of a globally-installed executable like we do with other hooks

@sds `commitplease` currently does not support global install, only local. I will be looking at that in jzaefferer/commitplease#74 but that really should not be a problem. The idea is to install `commitplease` on a per-developed-package basis and configure its options individually for each package. 

> Has this been tested? Does it work?

Now the answer is "yes" to both, here is how:

Build the gem with the local changes:
```
gem build overcommit.gemspec
gem install ./overcommit-0.34.2.gem
```

Now create a test project:
```
mkdir ../project && cd ../poject
git init && npm init --yes
npm i commitplease --save-dev
overcommit --install
overcommit --sign
```

Enable commitplease hook in the `.overcommit.yml` of the project:
```
CommitMsg:
  Commitplease:
    enabled: yes
```

Sign the changes to `.overcommit.yml`
```
overcommit --sign
```

Attempt to create an initial commit (expecting commitplease to complain about message):
```
git add package.json .overcommit.yml
git commit -m "init"
```

And it works: overcommit replays the message from commitplease that "init" does not conform to jQuery style. So try again with a conforming message:

```
git commit -m "Chore: initialise repository"
```

It works. PR updated and tested, please review the ruby script as I do not know ruby.

[1]: https://github.com/jzaefferer/commitplease/issues/72#issuecomment-236370515

PR #410, opened 2016-08-03
Added pre-commit hook to check for Windows line endings

harmjanblok:
Proposal to add a hook to verify the used line endings. This could also be extend to be a more generic line endings check, for example for projects who insist on CRLF instead of LF.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -592,7 +592,8 @@ PreCommit:
   LineEndings:
     description: 'Check line endings'
     enabled: false
-    representation: 'unix'
+    eol: 'lf'
+    # eol: 'crlf'
 
   XmlLint:
     enabled: false
--- a/config/default.yml
+++ b/config/default.yml
@@ -592,8 +592,7 @@ PreCommit:
   LineEndings:
     description: 'Check line endings'
     enabled: false
-    eol: 'lf'
-    # eol: 'crlf'
+    eol: "\n"
 
   XmlLint:
     enabled: false
--- a/config/default.yml
+++ b/config/default.yml
@@ -589,6 +589,11 @@ PreCommit:
     include:
       - '**/*.html'
 
+  LineEndings:
+    description: 'Check line endings'
+    enabled: false
+    representation: 'unix'
+
   XmlLint:
     enabled: false
     description: 'Analyze with xmllint'
--- a/lib/overcommit/hook/pre_commit/line_endings.rb
+++ b/lib/overcommit/hook/pre_commit/line_endings.rb
@@ -33,6 +33,7 @@ module Overcommit::Hook::PreCommit
 
       result.stdout.split("\0").map do |file_info|
         i, _w, _attr, path = file_info.split
+        next if i == 'l/-text' # ignore binary files
         next if i == "l/#{eol}"
         path
       end.compact
--- a/lib/overcommit/hook/pre_commit/line_endings.rb
+++ b/lib/overcommit/hook/pre_commit/line_endings.rb
@@ -4,11 +4,14 @@ module Overcommit::Hook::PreCommit
     def run
       messages = []
 
-      text_files.map do |file_name|
+      offending_files.map do |file_name|
         file = File.open(file_name)
         file.each_line do |line|
-          next if unix? && line =~ /\A((?!\r).)*\n\z/
-          next if windows? && line =~ /\A.*\r\n\z/
+          # remove configured line-ending
+          line.gsub!(/#{config['eol']}/, '')
+
+          # detect any left over line-ending characters
+          next unless line.end_with?("\n", "\r")
 
           messages << Overcommit::Hook::Message.new(
             :error,
@@ -24,19 +27,24 @@ module Overcommit::Hook::PreCommit
 
     private
 
-    def text_files
-      result = execute(%w[git grep -zIl \'\' --], args: applicable_files)
+    def offending_files
+      result = execute(%w[git ls-files --eol -z --], args: applicable_files)
       raise 'Unable to access git tree' unless result.success?
 
-      result.stdout.split("\0")
+      result.stdout.split("\0").map do |file_info|
+        i, _w, _attr, path = file_info.split
+        next if i == "l/#{eol}"
+        path
+      end.compact
     end
 
-    def unix?
-      config['eol'] == 'lf'
-    end
-
-    def windows?
-      config['eol'] == 'crlf'
+    def eol
+      @eol ||=  case config['eol']
+                when "\n"
+                  'lf'
+                when "\r\n"
+                  'crlf'
+                end
     end
   end
 end
--- a/lib/overcommit/hook/pre_commit/line_endings.rb
+++ b/lib/overcommit/hook/pre_commit/line_endings.rb
@@ -4,7 +4,7 @@ module Overcommit::Hook::PreCommit
     def run
       messages = []
 
-      applicable_files.map do |file_name|
+      text_files.map do |file_name|
         file = File.open(file_name)
         file.each_line do |line|
           next if unix? && line =~ /\A((?!\r).)*\n\z/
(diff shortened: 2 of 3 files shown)


jawshooah:
> This could also be extend to be a more generic line endings check

Let's do it that way. I don't think it makes sense for us to be opinionated here, since we support both *nix and Windows.

lencioni:
Great idea for a hook, thanks for bringing it!

I agree with @jawshooah. Also, it looks like the AppVeyor build failed, so you'll want to fix that up before we can merge this.

Thanks for working on this @harmjanblok

harmjanblok:
I've updated the commit to a more generic line ending check. The line ending representation can be configured to either `unix` or `windows`.

Also I've rebased my branch against latest master. The builds are still running, I'll keep an eye on them.

Is there some way to document the possible representations?

harmjanblok:
A check for excluding binary files is still missing. I'll add that next week.

sds:
Thanks for your work on this, @harmjanblok!

Is there any reason you didn't just use `grep` to perform these searches? `grep` comes with built-in support for filtering out binary files, and would reduce the amount of code you need to write here.

See the [`HardTabs` pre-commit hook](https://github.com/brigade/overcommit/blob/5509840ed5100cb64defc31ee96fb60fdbf174ed/lib/overcommit/hook/pre_commit/hard_tabs.rb) and [its corresponding configuration](https://github.com/brigade/overcommit/blob/5509840ed5100cb64defc31ee96fb60fdbf174ed/config/default.yml#L272-L277) for an example.

harmjanblok:
Yes @sds in my first approach in actually used `grep`, this first approach (`grep '\r$'`) only matched windows line endings on unix systems. The [tests failed on window systems](https://ci.appveyor.com/project/brigade/overcommit/build/153). I assume this approach failed on windows systems because:

> newline is also a separator for the list of patterns, there is no way
to match newline characters in a text. (from: [GNU grep Introduction](http://www.gnu.org/software/grep/manual/grep.html#Introduction))

I wanted to extend this first approach in two ways:
- use positive matching (`line =~ /.*\n/`) instead of negative matching (`!(line =~ /.*\r\n/)`)
- create a generic approach for both windows and unix systems to match the desired line endings

So I decided to switch to native `ruby` because of `grep` specific newline behavior, and because I couldn't find any documentation on the behavior of windows `grep` regarding line endings (for example does `grep` maybe convert `\n` to `\r\n` on windows?).

To verify the performance impact I created a [small benchmark](https://gist.github.com/harmjanblok/ca041bc44a9c54f325fed6f0abf7fdf5), where I compared two `ruby` implementations with `grep`. From this benchmark `ruby` seems to perform very well compared to `grep`.

The current solution however still uses negative matching, so I've refactored the used matches (see 04b6225).

But maybe these refactored matches can be used with `grep` as well, so I'll give `grep` one more try (in the next few days).

config/default.yml
```diff
@@ -589,6 +589,12 @@ PreCommit:
     include:
       - '**/*.html'
 
+  LineEndings:
+    description: 'Check line endings'
+    enabled: false
+    eol: 'lf'
+    # eol: 'crlf'
```
sds: It looks like you changed from `unix` / `windows` to `lf` / `crlf` in a subsequent commit. Any reason we shouldn't go with just listing the literal characters, i.e. `\n` and `\r\n`?

Looking at the list of possible [line ending representations](https://en.wikipedia.org/wiki/Newline#Representations), I'm skeptical we'll _ever_ have to support anything besides what are now called Unix and Windows line endings, but if we allow specifying the literal characters in the option then we're open to supporting any line ending.

lib/overcommit/hook/pre_commit/line_endings.rb
```diff
@@ -0,0 +1,42 @@
...
+  # Checks for line endings in files.
+  class LineEndings < Base
+    def run
+      messages = []
+
+      text_files.map do |file_name|
+        file = File.open(file_name)
+        file.each_line do |line|
+          next if unix? && line =~ /\A((?!\r).)*\n\z/
+          next if windows? && line =~ /\A.*\r\n\z/
```
sds: Perhaps there's a nuance here I'm missing, but is there any reason we can't use `String#end_with?` here? For example:

```ruby
next if unix? && !line.end_with?("\r\n")
next if windows? && line.end_with?("\r\n")
```

Performing a regex match for every line in a file will be much slower than running `end_with?`. For example:

```ruby
require 'benchmark'

iterations = 1_000_000
line = "this is a line\r\n"

Benchmark.bmbm do |x|
  x.report('regex') do
    iterations.times { line =~ /\A((?!\r).)*\n\z/ }
  end

  x.report('end_with?') do
    iterations.times { line.end_with?("\r\n") }
  end
end
```
```
Rehearsal ---------------------------------------------
regex       0.910000   0.000000   0.910000 (  0.912726)
end_with?   0.120000   0.000000   0.120000 (  0.120560)
------------------------------------ total: 1.030000sec

                user     system      total        real
regex       0.870000   0.000000   0.870000 (  0.875747)
end_with?   0.130000   0.000000   0.130000 (  0.126313)
```

These optimizations might not seem like much, but when you're running `overcommit --run` on a repository with many large files the performance will matter.

harmjanblok:
Thanks for your review @sds. Before going into detail on the current implementation, there's one more idea I came across and I would like to know your opinion on. Git itself can show line ending information for the files it's tracking with `git ls-files --eol` (lf, crlf or mixed). However this doesn't include the line-number(s) for lines with a non-default line ending.

config/default.yml
harmjanblok: I don't remember exactly why I switched to `lf`/`crlf`, but probably influenced by `git ls-files` as mentioned below. Yes going with literal characters would even be better. I'll add this to the next iteration.

lib/overcommit/hook/pre_commit/line_endings.rb
harmjanblok: I wanted to use positive matching (`line =~ /.*\n/`) instead of negative matching (`!(line =~ /.*\r\n/)`), but for performance it definitely makes more sense to use `end_with?`.

harmjanblok:
I added one more approach, with a `git ls-files` pre filter and a configurable line-ending detection mechanism. Let me know which solution you prefer.

lib/overcommit/hook/pre_commit/line_endings.rb
```diff
@@ -0,0 +1,51 @@
...
+  # Checks for line endings in files.
+  class LineEndings < Base
+    def run
+      messages = []
+
+      offending_files.map do |file_name|
+        file = File.open(file_name)
+        file.each_line do |line|
+          # remove configured line-ending
+          line.gsub!(/#{config['eol']}/, '')
```
sds: This will match all occurrences of `config['eol']` in the line, not just the last occurrence, no?

lib/overcommit/hook/pre_commit/line_endings.rb
```diff
@@ -0,0 +1,51 @@
...
+      end.compact
+    end
+
+    def eol
+      @eol ||=  case config['eol']
+                when "\n"
+                  'lf'
+                when "\r\n"
+                  'crlf'
+                end
```
sds: We should handle the case when the user specifies an unsupported line ending, rather than failing silently and return `nil`.

lib/overcommit/hook/pre_commit/line_endings.rb
sds: I guess `each_line` handles this for us, so nevermind. 😄

sds:
Looking at the implementation, it'll be easier if we stick to `\n` and `\r\n` as the allowed values for `eol`. 

Thanks for your hard work on this, @harmjanblok. I've merged this in f43743b with minor adjustments in 9101f32.

sds:
One thing worth mentioning: the `--eol` flag was added to `git ls-files` in Git 2.10, and as such is relatively new.

I've decided to not let that block this feature, and have modified the tests to only run for Git >= 2.10 in ddb4b5f.

harmjanblok:
@sds Thanks for finishing this feature!

PR #414, opened 2016-08-13
Add pre commit hook for fasterer

morizyun:
This hook is to run [fasterer](https://github.com/DamirSvrtan/fasterer) to suggest some speed improvements which you can check in detail at the [fast-ruby repo](https://github.com/JuanitoFatas/fast-ruby).

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -247,6 +247,11 @@ PreCommit:
     description: 'Check for file execute permissions'
     quiet: true
 
+  Fasterer:
+    enabled: false
+    description: 'Suggest some speed improvements'
+    required_executable: 'fasterer'
+
   ForbiddenBranches:
     enabled: false
     description: 'Check for commit to forbidden branch'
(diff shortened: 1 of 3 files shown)


morizyun:
@sds Thank you very much 👍

config/default.yml
```diff
@@ -247,6 +247,11 @@ PreCommit:
     description: 'Check for file execute permissions'
     quiet: true
 
+  Fasterer:
+    enabled: false
+    description: 'Suggest some speed improvements'
+    required_executable: 'fasterer'
```
jawshooah: Seems like this should have `include: '**/*.rb'` since it's Ruby-specific and doesn't require a full repo context like Brakeman does.

sds: Thanks for catching! Addressed in 4d9402b59fc51fde2af441c31175a9f5e1ab07c1.

morizyun: Thanks a lot!

PR #416, opened 2016-08-15
Return from protected? method unless find regex pattern

splybon:
This allows git push --tags to work.  Normally it can't find a ref_name


sds:
Thanks for the pull request @splybon!

I'm a little concerned about no test for this particular change. How did you test that this actually fixes your problem? It seems like `remote_ref` would be `nil` in the event no remote ref is provided, not an empty string. If it's `nil`, then `nil[%r{refs/heads/(.*)}, 1]` will result in a `NoMethodError`.

sds:
Closing due to lack of response. The fact that `NoMethodError` would appear to be raised here is my main concern. Happy to merge a PR with an accompanying test!

PR #418, opened 2016-08-18
Run fasterer only against changed Ruby files

jawshooah:


morizyun:
@jawshooah Thank you very much!

sds:
Thanks for your attention to detail on this, @jawshooah.

PR #429, opened 2016-09-09
Fix RailsSchemaUpToDate for new Rails projects

### LINKED ISSUE #417: RailsSchemaUpToDate fails on initial commit of a newly generated Rails app
1. Generate a new Rails app with `rails new`
2. In the generated app, run `rake db:migrate` as instructed by Rails
3. Initialize a git repo and install overcommit with `RailsSchemaUpToDate` enabled
4. Attempt to commit the project

RailsSchemaUpToDate halts the commit with this error:

```
Running pre-commit hooks
Check if database schema is up to date..........[RailsSchemaUpToDate] FAILED
You're trying to change the schema without adding a migration file
```

This is because a brand new Rails app has no migration files. But it *does* have a `schema.rb` (at version 0).

I believe RailsSchemaUpToDate should handle this case.

mattbrictson:
A new Rails app has no migration files, but if `db:migrate` has been run, then it will have a `schema.rb` with version 0. This would result in the failure "You're trying to change the schema without adding a migration file", which is a false negative.

Fix this by adding a `non_zero_schema_version?` test such that a schema at version 0 is allowed to be committed without any migrations present.

I've included a test and CHANGELOG entry. Fixes #417

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 * Add [`Brakeman`](http://brakemanscanner.org/) pre-push hook
 * Add [`TSLint`](http://palantir.github.io/tslint/) pre-commit hook
 * Validate that hook `env` environment configurations have valid names/values
+* Fix a false negative reported by RailsSchemaUpToDate for newly-created Rails
+  projects that don't yet have any migrations
 
 ## 0.35.0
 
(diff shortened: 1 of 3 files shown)


sds:
Thanks so much for this fix, @mattbrictson!

PR #434, opened 2016-09-27
Skip Overcommit builds for JRuby

sds:
Our Travis builds fail for JRuby with a mysterious:

```
  Check Travis CI configuration......................[TravisLint] FAILED
  Hook raised unexpected error
  Cannot run program "travis" (in directory "/home/travis/build/brigade/overcommit"):
  error=2, No such file or directory
```

...and similarly for other executables, e.g. `rubocop`.

This appears to be specific to the JRuby runtime. Since we don't care if
Overcommit runs work on JRuby, skip Overcommit for JRuby builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,8 @@ before_script:
 
 script:
   - bundle exec rspec
+  # Skip Overcommit on JRuby since it's unstable and doesn't actually test Overcommit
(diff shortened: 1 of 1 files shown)


jawshooah:
If we don't care whether Overcommit works with JRuby, why not remove the JRuby job entirely?

sds:
I still want to run the RSpec tests, I don't care about the result of `overcommit --run` in JRuby.

Probably could have worded the commit message differently.

PR #436, opened 2016-09-29
Add pre commit hook for Foodcritic

ajgon:


sds:
Thanks for submitting this pull request, @ajgon. I'm excited to integrate Foodcritic with Overcommit.

One potential gotcha I see here is that this integration doesn't leverage Foodcritic's support for linting environment and role directories. At @brigade, we actually use a monolithic repo to manage all of our Chef cookbooks (for a number of reasons, but one of them being we can give Foodcritic the context it needs to perform checks across multiple cookbooks, though I'm not sure if it actually does that yet).

We have a custom hook we use for this, which I'm happy to share below:

```ruby
module Overcommit::Hook::PreCommit
  # Lints cookbooks/environments/roles with `foodcritic`.
  class Foodcritic < Base
    def run
      repo_root = Overcommit::Utils.repo_root

      modified_cookbooks = directories_changed(File.join(repo_root, config['cookbooks_directory']))
      modified_envs = directories_changed(File.join(repo_root, config['environments_directory']))
      modified_roles = directories_changed(File.join(repo_root, config['roles_directory']))

      linter = ::FoodCritic::Linter.new
      review = linter.check(
        chef_version: Gem.loaded_specs['chef'].version.to_s,
        cookbook_paths: modified_cookbooks,
        environment_paths: modified_envs,
        role_paths: modified_roles,
        tags: @config['excluded_checks'].map { |check| "~#{check}" },
      )

      # Sort by rule code so we can easily fix offenses by type
      review.warnings.sort_by { |w| w.rule.code }.map do |warning|
        Overcommit::Hook::Message.new(
          warning.failed? ? :error : :warning,
          warning.match[:filename],
          warning.match[:line],
          "[#{warning.rule.code}] #{warning.rule.name}: " \
          "#{warning.match[:filename]}:#{warning.match[:line]}",
        )
      end
    end

    private

    def directories_changed(dir_prefix)
      applicable_files.
        select    { |path| path.start_with?(dir_prefix) }.
        map       { |path| path.gsub(%r{^#{dir_prefix}/}, '') }.
        group_by  { |path| path.split('/').first }.
        keys.
        map { |cookbook| File.join(dir_prefix, cookbook) }
    end
  end
end
```

Notice we integrate with the `Foodcritic::Linter` class directly. This is why we didn't contribute this upstream. Though it's not a hard requirement, it's better if Overcommit hooks interact with CLIs rather than Ruby APIs, as anecdotally CLIs are less-likely to break over time.

I'd love to see this pull request modified to support specifying additional flags like the Chef version, environment paths, role paths, etc. If you don't have time I may get to it myself, though not for a little bit. Thanks for getting the ball rolling!

ajgon:
Hey, I pushed the fixes, however I'm not a chef expert (never did roles or envs), so can you please check that this is actually what you asked for?

I tested in on my projects, and it seems to be working :)

ekrupnik:
+1

I agree that integrating Foodcritic into Overcommit would be a great enhancement. Thanks @ajgon.

lib/overcommit/hook/pre_commit/foodcritic.rb
```diff
@@ -0,0 +1,95 @@
...
+      @modified ||= {}
+      @modified[type] ||= directories_changed(full_directory_path("#{type}_directory"))
+    end
+
+    def full_directory_path(config_option)
+      return config[config_option] if config[config_option].start_with?(File::SEPARATOR)
+      File.absolute_path(File.join(repo_root, config[config_option]))
+    end
+
+    def repo_root
```
sds: This helper is used in one place, making it somewhat unnecessary given it is calling just another method (with no arguments/logic).

lib/overcommit/hook/pre_commit/foodcritic.rb
```diff
@@ -0,0 +1,95 @@
...
+      modified 'environments'
+    end
+
+    def modified_roles
+      modified 'roles'
+    end
+
+    def modified_cookbooks
+      modified 'cookbooks'
+    end
```
sds: These `modified_*` helpers are unnecessary since their name contains the same literal text as the arguments passed to the underlying method within.

sds:
Merged in 10a2242 with amendments in 0df0658.

lib/overcommit/hook/pre_commit/foodcritic.rb
```diff
@@ -0,0 +1,95 @@
...
+  #   enabled: true
+  #   cookbooks_directory: './cookbooks'
+  #   environments_directory: './environments'
+  #   roles_directory: './roles'
+  class Foodcritic < Base
+    def run
+      args = modified_environments_args + modified_roles_args + modified_cookbooks_args
+
+      args += applicable_files.reject do |file|
+        %w[spec test].any? { |dir| file.include?("#{File::SEPARATOR}#{dir}#{File::SEPARATOR}") }
```
ajgon: Yeah, I included that because I had a problem with `foodcritic` interpreting them as a separate cookbooks, and I couldn't resolve it easily. It appears, that `-B root_path` which you did was the solution I was looking for :)

PR #439, opened 2016-10-03
Add licence header enforcement

### LINKED ISSUE #438: Enforce license headers in source
I'd like to use overcommit to enforce license headers are present in source files. 

I don't see any existing pre commit hooks that support this. So i'd like to implement one myself. 
Wanted to check to make sure i'm not missing a way to do this currently before i implement.

sectioneight:
Fixes #438

Man, my Ruby is rusty. I don't know the best way to test this. I've
tried it out on my project manually.

Tested manually:

```
Updating signature for configuration file...
Running pre-commit hooks
Check YAML syntax........................................[YamlSyntax] OK
Check for trailing whitespace....................[TrailingWhitespace] OK
Validate JSON syntax.....................................[JsonSyntax] OK
Run LicenceHeader.....................................[LicenceHeader] FAILED
/Users/aiden/src/go/src/github.com/uber-go/uberfx/uberfx.go missing header contents from line 18 of LICENSE.txt: THE SOFTWARE.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## master
 
 * Add [`YAMLLint`](https://github.com/adrienverge/yamllint) pre-commit hook
+* Add `LicenceHeader` pre-commit enforcement to ensure open source projects
+  contain proper licence comments.
 
 ## 0.36.0
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -371,6 +371,7 @@ PreCommit:
     enabled: false
     include: '**/*.*'
     licence_file: 'LICENSE.txt'
+    description: 'Check source files for licence headers'
 
   LocalPathsInGemfile:
     enabled: false
--- a/config/default.yml
+++ b/config/default.yml
@@ -369,7 +369,6 @@ PreCommit:
 
   LicenceHeader:
     enabled: false
-    include: '**/*.*'
     licence_file: 'LICENSE.txt'
     description: 'Check source files for licence headers'
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -367,6 +367,11 @@ PreCommit:
     install_command: 'gem install json'
     include: '**/*.json'
 
+  LicenceHeader:
+    enabled: false
+    include: '**/*.*'
+    licence_file: 'LICENSE.txt'
+
   LocalPathsInGemfile:
     enabled: false
     description: 'Check for local paths in Gemfile'
--- a/spec/overcommit/hook/pre_commit/licence_header_spec.rb
+++ b/spec/overcommit/hook/pre_commit/licence_header_spec.rb
@@ -1,0 +1,29 @@
+require 'spec_helper'
+
+describe Overcommit::Hook::PreCommit::LicenceHeader do
+  let(:config)  { Overcommit::ConfigurationLoader.default_configuration }
+  let(:context) { double('context') }
+  subject { described_class.new(config, context) }
+
+  context 'when licence file is missing' do
+    before do
+      subject.stub(:licence_lines).and_raise(Errno::ENOENT)
+    end
+
+    it { should fail_hook }
+  end
+
+  context 'with headers in all applicable files' do
+    let(:licence) do
+      ['Hello World', 'Copyleft 2017']
+    end
+
+    before do
+      subject.stub(:applicable_files).and_return(%w[main.go])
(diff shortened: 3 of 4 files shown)


spec/overcommit/hook/pre_commit/licence_header_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    before do
+      subject.stub(:applicable_files).and_return(%w[main.go])
+      subject.stub(:licence_lines).and_return(licence)
+      subject.stub(:check_file).and_return(nil)
+    end
+
+    it { should pass }
+  end
+end
```
sectioneight: I'd really like an "integration" spec where I lay out a repo with a LICENCE.txt file and a few files with and without licence headers. Looks like most of the tests here are stubs/doubles/mocks. Perhaps I've just been writing Go for too long, but am I missing an easier way in Ruby/rspec to run this closer to how it's actually going to run? I uncovered a ton of bugs in manual testing due to the amount of mock/expect/stub/and_return etc.

spec/overcommit/hook/pre_commit/licence_header_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  let(:context) { double('context') }
+  subject { described_class.new(config, context) }
+
+  context 'when licence file is missing' do
+    before do
+      subject.stub(:licence_lines).and_raise(Errno::ENOENT)
+    end
+
+    it { should fail_hook }
+  end
```
sds: Can rewrite this to be an "actual" test without stubs by writing (untested):

```ruby
around do |example|
  repo do
    example.run
  end
end 

context 'when license file is missing' do
  it { should fail_hook }
end

context 'when license file exists' do
  let(:license_contents) { <<-LICENSE }
  Look at me
  I'm a license
  LICENSE

  let(:file) { 'some-file.txt' }

  before do
    File.open('LICENSE.txt', 'w') { |f| f.write(license_contents) }
    subject.stub(:applicable_files).and_return([file])
  end

  context 'when all files contain the license header' do
    before do
      File.open(file, 'w') { |f| f.puts(license_contents); f.write('And some text') }
    end
 
    it { should pass }
  end

  context 'when all files contain the license header' do
    before do
      File.open(file, 'w') { |f| f.write('Some text without a license') }
    end

    it { should fail_hook }
  end
end
```

We use stubbing in the Overcommit test suite when integrating with third-party tools, since we don't want to require developers to install those tools locally in order to run tests for Overcommit. In this case, creating actual files is totally fine.

sectioneight:
Awesome. Will address all comments and add some "real" tests. Thanks Shane! Also hello :)

config/default.yml
```diff
@@ -367,6 +367,12 @@ PreCommit:
     install_command: 'gem install json'
     include: '**/*.json'
 
+  LicenceHeader:
+    enabled: false
+    include: '**/*.*'
```
sectioneight: Nope, this was me not realizing there were sane defaults. Fixing.

spec/overcommit/hook/pre_commit/licence_header_spec.rb
sectioneight: Awesome, your code worked out of the box :)

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## master
 
 * Add [`YAMLLint`](https://github.com/adrienverge/yamllint) pre-commit hook
+* Add `LicenceHeader` pre-commit enforcement to ensure open source projects
```
sectioneight: Not sure if this is supposed to go above or below

sectioneight:
Seems like the tests fail on ruby <2.1 due to dependencies?

sds:
Merged in a5fcece81281a3fb95f6643aedbcbdc523acc882.

Yeah, looks like some upstream gem dependencies now explicitly require Ruby 2.1 or newer. Since Ruby 2.0 was EOLed in February 2016 I'm cool with dropping support for it now, and have done so in 999038d.

Thanks for the pull request! Nice to see you back 'round these parts. 😄

ajgon:
I have one question regarding this - should't we use the `License` word (with "s" instead of "c"). I know, that both forms are valid, but the `License` word became more or less industry standard, and maybe we should stick to it, to avoid confusion? I addressed this issue in https://github.com/brigade/overcommit/pull/444, you can either accept it or close it, depending on how you feel with it :)

PR #441, opened 2016-10-12
Keep net-http-persistent below 3.0 for Ruby 2.0 compatibility

jawshooah:
No need to drop support for 2.0 altogether, we just need to pin `net-http-persistent` below `3.0`.


jawshooah:
Eh, its only dependent is `travis`, which would be no great loss if we had to drop it.

PR #444, opened 2016-10-13
Use `License` word instead of `Licence`

ajgon:
In `LicenceHeader` hook, we have a word `Licence`, but we should use
`License`. Both forms are valid, but the `License` word became more or
less industry standard, and we should stick to it, to avoid confusion.


sds:
No objections from my side! Pinging @sectioneight as he'll likely want to be made aware of the change. 👍

sds:
Merged in 80766b4. Thanks!

PR #455, opened 2016-12-27
Moved `Commitplease` hook from `CommitMsg` to `PostCommit`

ajgon:
`CommitMsg` hook works only on commit messages. Unfortunately
`commitplease` does not allow to pass a message directly on it, it only
works on posted commits. That's why this hooks needs to be moved to
`PostCommit` section - `commitplease` needs to actually see the
commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### New Features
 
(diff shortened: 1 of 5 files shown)


trotzig:
Thanks @ajgon! This makes sense, although it's a little unfortunate that commitplease can't deal with the commit message in isolation (commit-msg hooks get the commit message as a temp file).

PR #456, opened 2017-01-02
Add RakeTarget pre-commit and pre-push hook

vinzent:
Allow to run rake targets as pre-commit and pre-push hooks.

The first unsuccessfull exit code of a specified rake task
will fail the hook.

Example configuration: 
 
```yaml 
PreCommit: 
  RakeTask: 
    enabled: true 
    targets: 
      - 'validate' 
      - 'lint' 
 
PrePush: 
  RakeTask: 
    enabled: true 
    targets: 
      - 'test' 
```

If considired for merging I would require  a hint howto test for a raised error on missing targets parameter. :-)

### DIFF
--- a/lib/overcommit/hook/shared/rake_target.rb
+++ b/lib/overcommit/hook/shared/rake_target.rb
@@ -1,0 +1,24 @@
+module Overcommit::Hook::Shared
+  # runs specified rake targets. It fails on the first non-
+  # successfull exit.
+  #
+  module RakeTarget
+    def run
+      targets = config['targets']
+
+      if Array(targets).empty?
+        raise 'RakeTarget: targets parameter is empty. Add at least one task to ' \
+          'the targets parameter. Valid: Array of target names or String of ' \
+          'target names'
+      end
+
+      targets.each do |task|
+        result = execute(command + [task])
+        unless result.success?
+          return :fail, "Rake target #{task}:\n#{result.stdout}"
+        end
+      end
+      :pass
+    end
+  end
+end
--- a/spec/overcommit/hook/pre_commit/rake_target_spec.rb
+++ b/spec/overcommit/hook/pre_commit/rake_target_spec.rb
@@ -1,0 +1,49 @@
+require 'spec_helper'
+
+describe Overcommit::Hook::PreCommit::RakeTarget do
+  let(:config)  { Overcommit::ConfigurationLoader.default_configuration }
+  let(:context) { double('context') }
+  subject { described_class.new(config, context) }
+
+  context 'without targets parameters' do
+    let(:result) { double('result') }
+    it 'raises' do
+      expect { subject.run }.to raise_error(
+        RuntimeError, /RakeTarget: targets parameter is empty.*/
+      )
+    end
+  end
+
+  context 'with targets parameter set' do
+    let(:config) do
+      super().merge(Overcommit::Configuration.new(
+        'PreCommit' => {
+          'RakeTarget' => {
+            'targets' => ['test'],
+          }
+        }
+      ))
+    end
+    let(:result) { double('result') }
+
+    context 'when rake exits successfully' do
+      before do
+        result.stub(:success?).and_return(true)
+        subject.stub(:execute).and_return(result)
+        result.stub(:stdout).and_return('ANYTHING')
+      end
+
+      it { should pass }
+    end
+
+    context 'when rake exits unsuccessfully' do
+      before do
+        result.stub(:success?).and_return(false)
+        subject.stub(:execute).and_return(result)
+        result.stub(:stdout).and_return('ANYTHING')
+      end
+
+      it { should fail_hook }
+    end
+  end
+end
--- a/spec/overcommit/hook/pre_push/rake_target_spec.rb
+++ b/spec/overcommit/hook/pre_push/rake_target_spec.rb
@@ -1,0 +1,49 @@
+require 'spec_helper'
+
+describe Overcommit::Hook::PrePush::RakeTarget do
+  let(:config)  { Overcommit::ConfigurationLoader.default_configuration }
+  let(:context) { double('context') }
+  subject { described_class.new(config, context) }
+
+  context 'without targets parameters' do
+    let(:result) { double('result') }
+    it 'raises' do
+      expect { subject.run }.to raise_error(
+        RuntimeError, /RakeTarget: targets parameter is empty.*/
+      )
+    end
+  end
+
+  context 'with targets parameter set' do
+    let(:config) do
+      super().merge(Overcommit::Configuration.new(
+        'PrePush' => {
+          'RakeTarget' => {
+            'targets' => ['test'],
+          }
+        }
+      ))
+    end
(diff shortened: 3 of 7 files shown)


lib/overcommit/hook/shared/rake_target.rb
```diff
@@ -0,0 +1,24 @@
+module Overcommit::Hook::Shared
+  # runs specified rake targets. It fails on the first non-
+  # successfull exit.
+  #
+  module RakeTarget
+    def run(targets = config['targets'])
```
sds: It's a bit odd to define a parameter list here when the `Overcommit::Hook::Base#run` method does not define any parameters. You might as well pull this down into the method implementation.

spec/overcommit/hook/pre_commit/rake_target_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+describe Overcommit::Hook::PreCommit::RakeTarget do
+  let(:config)  { Overcommit::ConfigurationLoader.default_configuration }
+  let(:context) { double('context') }
+  subject { described_class.new(config, context) }
+
+  # context 'without targets parameters' do
+  #   let(:result) { double('result') }
+  #   it 'raises' do
+  #     expect { subject }.to raise_error(RuntimeError)
```
sds: If you want to test the error case, you can check the content of the error message rather than the type of error using RSpec's regex matcher, e.g.

```ruby
expect { subject }.to raise_error(/some message/i)
```

spec/overcommit/hook/pre_commit/rake_target_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  #   end
+  # end
+
+  context 'wth targets parameter set' do
```
sds: "wth" -> "with"

spec/overcommit/hook/pre_push/rake_target_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  subject { described_class.new(config, context) }
+
+  # context 'without targets parameters' do
+  #   let(:result) { double('result') }
+  #   it 'raises' do
+  #     expect { subject }.to raise_error(RuntimeError)
+  #   end
+  # end
+
+  context 'wth targets parameter set' do
```
sds: "wth" -> "with"

lib/overcommit/hook/shared/rake_target.rb
```diff
@@ -0,0 +1,24 @@
...
+      if targets.nil?
+        raise "RakeTarget: targets parameter is nil. Add at least one task to \
+          the targets parameter. Valid: Array of target names"
+      elsif !targets.is_a?(Array)
+        raise "RakeTarget: targets parameter is not an array. Valid: Array of \
+          rake target names"
+      end
```
sds: You could simplify this into a single case by `Array`ifying `targets`:

```ruby
if Array(targets).empty?
  raise ...
end
```

This also allows users to specify a single string for `targets` so they don't need to explicitly make it an array, making it slightly more user-friendly. Overcommit takes this approach with other configuration options where it makes sense.

lib/overcommit/hook/shared/rake_target.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      targets.each do |task|
+        result = execute(command << task)
```
sds: Fortunately the `Overcommit::Hook::Base#command` always returns a new array, but I don't believe we use this style of appending a single item using `<<` elsewhere in the code where performance isn't key (preferring a more functional approach), since it can potentially cause problems _if_ the method were to return a reference to a memoized value, causing each invocation to modify the list of command arguments.

Would prefer if we changed it to:

```ruby
  ... = execute(command + [task])
```

...as the overhead of creating a new array in this case isn't worth worrying about.

lib/overcommit/hook/shared/rake_target.rb
vinzent: adressed in FIXUP commit 2cf4462da54100be753f1e334b0e01ca03e5e457

lib/overcommit/hook/shared/rake_target.rb
vinzent: addressed in FIXUP commit 9789302ccd394e340451e7d673c8f771ee67d6e2

lib/overcommit/hook/shared/rake_target.rb
vinzent: adressed in FIXUP commit 37aa85a2c1eee1cdd5f990b817f006db497649c9

spec/overcommit/hook/pre_commit/rake_target_spec.rb
vinzent: adressed in FIXUP commit 6d28647d6173a6d4a79950a06fb2a80edb5f4cc8

spec/overcommit/hook/pre_push/rake_target_spec.rb
vinzent: adressed in FIXUP commit 6d28647d6173a6d4a79950a06fb2a80edb5f4cc8

spec/overcommit/hook/pre_commit/rake_target_spec.rb
vinzent: this fails for me (added a message checkt too):

```
    
    Failure/Error: expect { subject }.to raise_error(RuntimeError,
      /RakeTarget: targets parameter is empty.*/)
      expected RuntimeError with message matching /RakeTarget:
      targets parameter is empty.*/ but nothing was raised
```

(f4a84b4b4c993ed32c5fdfe5e964283707c23dda)

vinzent: but if testing with a real example:

```
Hook raised unexpected error
RakeTarget: targets parameter is empty. Add at least one task to           the targets parameter. Valid: Array of target names or String of           target names
```

and it seems like i need to break lines in another way in the code.

vinzent:
@sds thanks for your review!

spec/overcommit/hook/pre_commit/rake_target_spec.rb
```diff
@@ -5,12 +5,12 @@
...
-  # context 'without targets parameters' do
-  #   let(:result) { double('result') }
-  #   it 'raises' do
-  #     expect { subject }.to raise_error(RuntimeError)
-  #   end
-  # end
+  context 'without targets parameters' do
+    let(:result) { double('result') }
+    it 'raises' do
+      expect { subject }.to raise_error(RuntimeError, /RakeTarget: targets parameter is empty.*/)
```
trotzig: You need to call `subject.run` here in order to get the error thrown. This is done automatically in `spec/support/matchers/hook.rb` when you call `fail_hook` or `pass`. But since you're not using those helpers here you need to be more explicit.

vinzent:
squashed all FIXUPs.

trotzig:
Thanks @vinzent!

PR #459, opened 2017-01-09
Added option to run rspec only over modified files

simoneromanelli:
Hi!
I thought to add the option to run rspec only over the files modified in the current branch. In my personal experience have to wait for all the tests running was to much so I implemented this custom hook for my company project. I thought that maybe you could be interested in adding this feature to the rspec hook. 
Let me know what do you think. 
Best regards.
Simone.

### DIFF
--- a/lib/overcommit/git_repo.rb
+++ b/lib/overcommit/git_repo.rb
@@ -272,5 +272,23 @@ module Overcommit
     def current_branch
       `git symbolic-ref --short -q HEAD`.chomp
     end
+
+    # Returns the names of the files that have been modified comparing the
+    # current local branch with the corresponding remote one if it does exist or
+    # the current local branch with the parent one if not.
+    # @param remote [String] name of the remote
+    #
+    # @return [Array<String>] list of relative file paths
+    def modified_file_in_branch(remote)
+      remote_url = `git remote get-url #{remote}`.split("\n").first
+      parent_branch = `git show-branch | grep '*' | grep -v "#{current_branch}" | head -n1`.
+                      match(/\[([^~^\]]+)/).captures.first
+      commits_count = if `git ls-remote --heads #{remote_url} #{current_branch}`.empty?
+                        `git rev-list #{parent_branch}..#{current_branch} --count`
+                      else
+                        `git rev-list #{remote}/#{current_branch}..#{current_branch} --count`
+                      end
+      `git diff --name-only HEAD~#{commits_count}`.split("\n")
+    end
   end
 end
--- a/lib/overcommit/hook/pre_push/r_spec.rb
+++ b/lib/overcommit/hook/pre_push/r_spec.rb
@@ -1,14 +1,47 @@
 module Overcommit::Hook::PrePush
   # Runs `rspec` test suite before push
+  # .overcommit.yml configuration:
+  # Set `changed: true` to run rspec only over edited files.
+  # Set `remote`: $remote_name` to specify a remote name.
   #
   # @see http://rspec.info/
   class RSpec < Base
     def run
-      result = execute(command)
+      command = ['rspec']
+      remote = config['remote'] || 'origin'
+      result = if config['changed']
+                 return :pass if testable_files(remote).empty?
+                 execute([*command, *testable_files(remote)])
+               else
+                 execute(command)
+               end
       return :pass if result.success?
 
       output = result.stdout + result.stderr
       [:fail, output]
+    end
+
+    def testable_files(remote)
+      return [] if Overcommit::GitRepo.modified_file_in_branch(remote).empty?
+      files = Overcommit::GitRepo.modified_file_in_branch(remote).map do |file_path|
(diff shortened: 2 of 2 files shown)


lib/overcommit/hook/pre_push/r_spec.rb
```diff
@@ -1,14 +1,47 @@
...
+    def test?(file_path)
+      file_path.start_with?('spec/') && file_path.end_with?('_spec.rb')
+    end
+
+    def ruby_file?(file_path)
+      file_path.end_with? '.rb'
+    end
+
+    def spec_file(file_path)
+      spec_path = "spec/#{file_path.insert(-4, '_spec')}".delete('/app')
```
sds: `-4` is a big of a magic number. We should base it off of the length of the extension, e.g. `'.rb'.length`.

This code also appears to assume the developer is using a Rails project (from the presence of the `/app` directory). Furthermore, `delete('/app')` is dangerous because it doesn't just delete the first occurrence, e.g.

```ruby
irb(main):001:0> "/app/hello/app-directory".delete("/app")
=> "hello-directory"
```

I don't know the best way to accomplish this, but it might be worth creating a mechanism of defining a "projection". Perhaps using a regex with capture groups to transform source files into their corresponding test files is an approach worth exploring.

Also, you'll want to run tests when the actual test files change. This change appears to only run corresponding test files for changed source files.

lib/overcommit/hook/pre_push/r_spec.rb
```diff
@@ -1,14 +1,47 @@
...
+      files = Overcommit::GitRepo.modified_file_in_branch(remote).map do |file_path|
+        if test?(file_path) || ruby_file?(file_path)
+          test?(file_path) ? file_path : spec_file(file_path)
+        end
+      end
+      files.uniq.reject { |file| file.nil? || file == false }
+    end
+
+    def test?(file_path)
+      file_path.start_with?('spec/') && file_path.end_with?('_spec.rb')
```
sds: I'm not sure if RSpec explicitly requires all specs be in the `spec` directory (although it is definitely a convention). Perhaps we should offer a way for the user to specify a list of patterns/globs which determine whether or not a file is a test file.

lib/overcommit/hook/pre_push/r_spec.rb
```diff
@@ -1,14 +1,47 @@
...
   #
   # @see http://rspec.info/
   class RSpec < Base
     def run
-      result = execute(command)
+      command = ['rspec']
+      remote = config['remote'] || 'origin'
+      result = if config['changed']
+                 return :pass if testable_files(remote).empty?
+                 execute([*command, *testable_files(remote)])
```
sds: `testable_files` is called twice here; we should memoize the first call so we aren't recalculating unnecessarily.

lib/overcommit/hook/pre_push/r_spec.rb
```diff
@@ -1,14 +1,47 @@
...
   # Runs `rspec` test suite before push
+  # .overcommit.yml configuration:
+  # Set `changed: true` to run rspec only over edited files.
+  # Set `remote`: $remote_name` to specify a remote name.
   #
   # @see http://rspec.info/
   class RSpec < Base
     def run
-      result = execute(command)
+      command = ['rspec']
```
sds: You don't want to explicitly override `command` here with a local variable, as this breaks your ability to configure additional flags via the `flags` hook option.

lib/overcommit/git_repo.rb
```diff
@@ -272,5 +272,23 @@ def branches_containing_commit(commit_ref)
...
     end
+
+    # Returns the names of the files that have been modified comparing the
+    # current local branch with the corresponding remote one if it does exist or
+    # the current local branch with the parent one if not.
+    # @param remote [String] name of the remote
+    #
+    # @return [Array<String>] list of relative file paths
+    def modified_file_in_branch(remote)
+      remote_url = `git remote get-url #{remote}`.split("\n").first
```
sds: There's a danger here with using backticks as any failures are silently ignored (potentially returning an empty string). 

This isn't a big deal but if you have time it might be worth using the `execute` helper so you can actually inspect the exit status of the process. Not a blocker.

simoneromanelli:
Hi @sds 
Thanks for your fast feedback.
I'll do the possible to improve my code according with your suggestions. 
Cheers.

lib/overcommit/hook/pre_push/r_spec.rb
ryanfox1985: Good idea.

ryanfox1985: @simoneromanelli maybe you can add some heuristic algorithm to detects the tests related with the modified file. Example:
I modified `app/models/cars.rb` you can search the word `car` in the Rspec test files content.

sds:
Closing due to lack of activity. Feel free to reopen with a pull request addressing those comments!

PR #460, opened 2017-01-10
Add check to make sure ESLint has not failed to run

### LINKED ISSUE #458: Overcommit shows passing even when ESLint fails to execute correctly
We were recently having an issue where ESLint was always showing in overcommit as passing on travis even though our version of node was too old and thus ESLint was failing to run (due to the fact that an incompatible version of node came with travis). 

The output of `eslint .` can be seen [here](https://travis-ci.org/projectcypress/cypress/builds/189308181#L532) and the output of overcommit on the same run can be seen [here](https://travis-ci.org/projectcypress/cypress/builds/189308181#L540).

 In cases where ESLint fails to execute correctly, it seems that it would be better to just have ESLint fail other than blindly pass. In this case it meant we were able to merge things into production that were not actually passing our linters.

rbclark:
Previously if the eslint binary fails to run overcommit will simply show all ESLint tests as passing.

This checks to see if eslint has written anything to standard error before simply assuming everything is okay.

Closes #458

### DIFF
--- a/lib/overcommit/hook/pre_commit/es_lint.rb
+++ b/lib/overcommit/hook/pre_commit/es_lint.rb
@@ -21,6 +21,8 @@ module Overcommit::Hook::PreCommit
       output = result.stdout.chomp
       return :pass if result.success? && output.empty?
 
+      return [:fail, result.stderr] unless result.stderr.empty?
+
       # example message:
       #   path/to/file.js: line 1, col 0, Error - Error message (ruleName)
       extract_messages(


trotzig:
Thanks for taking a stab at this @rbclark. I think we're not getting to the root cause however. The fix you have will break the check when eslint outputs a warning. Would you mind sharing the .overcommit.yml configuration for the EsLint hook? I'm trying to reproduce this locally but I can't make overcommit pass when the eslint check fails.

rbclark:
Oops, yep I thought I had it working right locally but I guess not 🙁 , I am using https://github.com/projectcypress/cypress/blob/master/.overcommit.yml for the overcommit configuration and have setup a Ubuntu 12.04 test machine running nodenv with version `0.10.48` of node installed and the latest eslint. Running `eslint .` in this scenario causes eslint to error out for me.

trotzig:
Hey, sorry I didn't respond to you earlier. Could you try with a more recent node version. It looks like [eslint only supports Node v4 and higher](https://github.com/eslint/eslint/blob/master/.travis.yml).

sds:
Closing due to lack of activity. Feel free to reopen with those comments addressed!

PR #467, opened 2017-01-26
Fix jruby build

trotzig:
This is the result of me exploring why the jruby build suddenly started failing. An update to bundler made installing the rainbow gem fail. 

https://github.com/sickill/rainbow/issues/48

I kept all the attempts I made at fixing the issue as separate attempts, to help understand the process better. I'll probably squash before merging though, to keep the git history a little cleaner.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_script:
   - git config --global user.name "Travis CI"
 
 before_install:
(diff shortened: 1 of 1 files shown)


trotzig:
Sigh, not sure what's wrong with the appveyor build. I only made changes in `.travis.yml`, so I'm assuming the spec failure is spurious.

sds:
Agree the failure on Appveyor appears unrelated. Thanks for fixing!

PR #468, opened 2017-01-30
Make HookContext::PreCommit#initial_commit? public

thegranddesign:
Since it's being forwarded to, in Ruby 2.4 it throws a warning.


sds:
Thanks for the pull request! The test failures were unrelated.

PR #472, opened 2017-03-01
Auto-sign config file when installing hooks the first time

sds:
Previously, we required users to explicitly run `overcommit --sign` in their repositories after running `overcommit --install`. This was slightly annoying and made for a poor user onboarding experience.

For all practical purposes, the initial bootstrap case seems like one where ease of getting started with the tool trumps the risk of malicious code execution via hooks. It is reasonable to assume that a user who runs `overcommit --install` is accepting responsibility for hook code executed as of that point. It is only on subsequent changes to configuration/hooks that they should be warned to sign the config file/hooks appropriately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fix `forwarding to private method` warning on Ruby 2.4.x
 * Update `childprocess` dependency to 0.6.x series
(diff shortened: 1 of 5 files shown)


sds:
Windows builds were unrelated (we caused by being unable to install a specific gem `rainbows`). Since they passed on other Ruby versions I went ahead and merged.

PR #475, opened 2017-03-10
ESLint pre-commit hook fails if there are problems

domcorvasce:
A user can choose to ignore this behavior by specifying `bindly_pass` option:

```yaml
PreCommit:
  EsLint:
    enabled: true
    on_warn: fail
    bindly_pass: true
```

### DIFF
--- a/lib/overcommit/hook/pre_commit/es_lint.rb
+++ b/lib/overcommit/hook/pre_commit/es_lint.rb
@@ -12,9 +12,6 @@ module Overcommit::Hook::PreCommit
   #     enabled: true
   #     command: ['npm', 'run', 'lint', '--', '-f', 'compact']
   #
-  # If ESLint returns some errors, you can bindly pass by
-  # setting `bindly_pass` option to true.
-  #
   # Note: This hook supports only compact format.
   #
   # @see http://eslint.org/
@@ -24,11 +21,7 @@ module Overcommit::Hook::PreCommit
       output = result.stdout.chomp
       messages = output.split("\n").grep(/Warning|Error/)
 
-      # Fail for issues relative to the tool used.
-      if !config['bindly_pass'] && messages.empty? && !result.success?
-        return [:fail, result.stderr]
-      end
-
+      return [:fail, result.stderr] if messages.empty? && !result.success?
       return :pass if result.success? && output.empty?
 
       # example message:
--- a/lib/overcommit/hook/pre_commit/es_lint.rb
+++ b/lib/overcommit/hook/pre_commit/es_lint.rb
@@ -12,6 +12,9 @@ module Overcommit::Hook::PreCommit
   #     enabled: true
   #     command: ['npm', 'run', 'lint', '--', '-f', 'compact']
   #
+  # If ESLint returns some errors, you can bindly pass by
+  # setting `bindly_pass` option to true.
+  #
   # Note: This hook supports only compact format.
   #
   # @see http://eslint.org/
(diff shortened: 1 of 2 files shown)


lib/overcommit/hook/pre_commit/es_lint.rb
```diff
@@ -12,19 +12,30 @@ module Overcommit::Hook::PreCommit
   #     enabled: true
   #     command: ['npm', 'run', 'lint', '--', '-f', 'compact']
   #
+  # If ESLint returns some errors, you can bindly pass by
+  # setting `bindly_pass` option to true.
```
sds: Can we elaborate on the conditions where this is useful? (concrete examples are very helpful here)

lib/overcommit/hook/pre_commit/es_lint.rb
```diff
@@ -12,19 +12,30 @@ module Overcommit::Hook::PreCommit
...
   # @see http://eslint.org/
   class EsLint < Base
     def run
       result = execute(command, args: applicable_files)
       output = result.stdout.chomp
+      messages = output.split("\n").grep(/Warning|Error/)
+
+      # Fail for issues relative to the tool used.
+      if !config['bindly_pass'] && messages.empty? && !result.success?
+        $stderr.puts result.stderrs
```
sds: Using `$stderr.puts` will break the formatting of the hook run. Output should be returned along with the `return` line below, e.g.:

```ruby
return [:fail, result.stderr]
```

Also, `stderrs` is a typo (should be `stderr`).

spec/overcommit/hook/pre_commit/es_lint_spec.rb
```diff
@@ -9,6 +9,20 @@
     subject.stub(:applicable_files).and_return(%w[file1.js file2.js])
   end
 
+  context 'when eslint is unable to run' do
+    let(:result) { double('result') }
+
+    before do
+      result.stub(:stderrs).and_return('SyntaxError: Use of const in strict mode.')
```
sds: `stderrs` -> `stderr`

lib/overcommit/hook/pre_commit/es_lint.rb
domcorvasce: Honestly, I don't think it can be useful.
But I thought that some people might want to pass ahead if something goes wrong with ESLint (I won't do it).

lib/overcommit/hook/pre_commit/es_lint.rb
domcorvasce: I am very sorry for this.
I returned the stderrs instance variable.

spec/overcommit/hook/pre_commit/es_lint_spec.rb
domcorvasce: Mea culpa. I'll fix ASAP.
Thanks for your time.

domcorvasce:
Hello @sds!

First, thanks for the time you dedicate to this project and to me.
For the testing issue: I tested the changes locally and it works correctly, I tried multiple times before to open a pull request.

I am sorry 'cause I've not studied the code enough to avoid the typo I did.

lib/overcommit/hook/pre_commit/es_lint.rb
trotzig: In that case, I think it's better to remove it completely. If not having the configuration option turns out to be an issue for people, we can always add it at that point.

trotzig:
Thanks for following up @domcorvasce! The build failures weren't introduced by you so I went ahead and merged your changes.

PR #479, opened 2017-03-23
Add Git LFS pre-push hook

detouched:
Git LFS installs pre-push hook which:
- checks that git-lfs utility is on PATH, and if no, warns that Git LFS hook should probably be removed
- calls it by executing `git lfs pre-push <remote_name> <remote_url>`

The introduced hook is a Ruby port of the same algorithm
modulo hook disabling note in the warning message.

For the reference, here's the original hook script:

```sh
#!/bin/sh
command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting .git/hooks/pre-push.\n"; exit 2; }
git lfs pre-push "$@"
```

### DIFF
--- a/lib/overcommit/hook/pre_push/git_lfs.rb
+++ b/lib/overcommit/hook/pre_push/git_lfs.rb
@@ -1,0 +1,21 @@
+module Overcommit::Hook::PrePush
+  # Invokes Git LFS command that uploads files tracked by Git LFS to the LFS storage
+  #
+  # @see https://git-lfs.github.com/
+  class GitLfs < Base
+    def run
+      result = execute(['command', '-v', 'git-lfs'])
+      unless result.success?
+        return :warn, 'This repository is configured for Git LFS but \'git-lfs\' ' \
+        'was not found on your path.\nIf you no longer wish to use Git LFS, ' \
+        'disable this hook by removing or setting \'enabled: false\' for GitLFS ' \
+        'hook in your .overcommit.yml file'
+      end
+
+      result = execute(['git', 'lfs', 'pre-push', remote_name, remote_url])
+      return :fail, result.stderr unless result.success?
+
+      :pass
+    end
+  end
+end
--- a/lib/overcommit/hook/pre_push/git_lfs.rb
+++ b/lib/overcommit/hook/pre_push/git_lfs.rb
@@ -4,10 +4,9 @@ module Overcommit::Hook::PrePush
   # @see https://git-lfs.github.com/
   class GitLfs < Base
     def run
-      result = execute(['command', '-v', 'git-lfs'])
-      unless result.success?
+      unless in_path?('git-lfs')
         return :warn, 'This repository is configured for Git LFS but \'git-lfs\' ' \
-        'was not found on your path.\nIf you no longer wish to use Git LFS, ' \
+        "was not found on your path.\nIf you no longer wish to use Git LFS, " \
         'disable this hook by removing or setting \'enabled: false\' for GitLFS ' \
         'hook in your .overcommit.yml file'
       end
--- a/README.md
+++ b/README.md
@@ -549,6 +549,7 @@ but before any objects have been transferred. If a hook fails, the push is
 aborted.
 
 * [Brakeman](lib/overcommit/hook/pre_push/brakeman.rb)
+* [GitLfs](lib/overcommit/hook/pre_push/git_lfs.rb)
 * [Minitest](lib/overcommit/hook/pre_push/minitest.rb)
 * [ProtectedBranches](lib/overcommit/hook/pre_push/protected_branches.rb)
 * [Pytest](lib/overcommit/hook/pre_push/pytest.rb)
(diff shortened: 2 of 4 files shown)


lib/overcommit/hook/pre_push/git_lfs.rb
```diff
@@ -0,0 +1,21 @@
...
+        return :warn, 'This repository is configured for Git LFS but \'git-lfs\' ' \
+        'was not found on your path.\nIf you no longer wish to use Git LFS, ' \
+        'disable this hook by removing or setting \'enabled: false\' for GitLFS ' \
+        'hook in your .overcommit.yml file'
+      end
+
+      result = execute(['git', 'lfs', 'pre-push', remote_name, remote_url])
+      return :fail, result.stderr unless result.success?
+
+      :pass
```
detouched: Not sure: is it possible to output some information in case the hook passes?

In fact, ideally it'd be cool to be able to output something _while the hook works_ cause uploading LFS files might take a while. The original LFS hook reports status interactively, something like:
```sh
Git LFS: (0 of 1 files) 3.80 MB / 6.85 MB
```

detouched:
Hey @sds, how do I trigger AppVeyor build again? Looks like its failure is unrelated.

Also I'd love to get your feedback and get this merged eventually.

Thanks!

lib/overcommit/hook/pre_push/git_lfs.rb
```diff
@@ -0,0 +1,21 @@
+module Overcommit::Hook::PrePush
+  # Invokes Git LFS command that uploads files tracked by Git LFS to the LFS storage
+  #
+  # @see https://git-lfs.github.com/
+  class GitLfs < Base
+    def run
+      result = execute(['command', '-v', 'git-lfs'])
```
trotzig: Is it always safe to assume that a non-zero exit code means `git-lfs` wasn't installed? We could run a `which git-lfs` first for better precision.

lib/overcommit/hook/pre_push/git_lfs.rb
trotzig: I don't know of a good way. Would it make sense to instead of actually uploading just check if there is something to upload? The overcommit hook would then output something like 

```
Checking for files that need to be uploaded..........................[GitLFS] FAILED
./some-file.large hasn't been uploaded. Run `git lfs pre-push ... something` to upload. 
```

I'm not a lfs user myself, so this idea might not be possible at all. Or it might just be annoying to have to run the upload manually.

lib/overcommit/hook/pre_push/git_lfs.rb
sds: We have a helper for just this purpose: [`in_path?`](https://github.com/brigade/overcommit/blob/336c495d23befc7139810f2d9d58e42f8494313e/lib/overcommit/hook/base.rb#L89)

lib/overcommit/hook/pre_push/git_lfs.rb
sds: This would require some refactoring of the display logic for Overcommit. In the past we've kicked around the idea of supporting progress bars or similar, but this would require a much more complicated display controller that utilized more advanced cursor features so you could update multiple lines at the same time (since hooks run in parallel).

sds: Definitely open to the idea, but I think it's well outside the scope of what you're trying to accomplish with this change.

sds:
Overall looks good, @detouched. Let's switch to using `in_path?` and we can ignore the Appveyor build, which is notoriously flakey.

lib/overcommit/hook/pre_push/git_lfs.rb
detouched: > Would it make sense to instead of actually uploading just check if there is something to upload?

Not really, you'd just want it to be uploaded just with `git push`, without additionally calling `git lfs ...`. Also, I'd not make overcommit behaviour different from standard (meaning what is installed by LFS client by default).

> It's well outside the scope of what you're trying to accomplish with this change.

I totally agree, I was wondering if there's a way to do that, but if not, it's fine. It might be just mentioned in the doc since GitLfs hook is disabled by default and requires the user to add it in the config yaml.

lib/overcommit/hook/pre_push/git_lfs.rb
detouched: I simply followed what the original hook does, for consistency (original hook is installed by Git LFS client). I mentioned that [before](https://github.com/brigade/overcommit/pull/479#issue-216310541), here's its entire content:

```bash
#!/bin/sh
command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting .git/hooks/pre-push.\n"; exit 2; }
git lfs pre-push "$@"
```

But yeah, `in_path` looks like a safe replacement, will update.

detouched:
As discussed, changed the way LFS binary is looked up to `in_path` tool.

Issue #480, opened 2017-03-23
pre-push hook fails when pushing tags


hasghari:
When I run `git push --tags git@heroku.com:repo.git` the pre-push hook fails with the following exception:

```
Running pre-push hooks
Check for illegal pushes to protected branches....[ProtectedBranches] FAILED
Hook raised unexpected error
no implicit conversion of nil into String
/Users/hasghari/.rvm/gems/ruby-2.2.5@ue/gems/overcommit-0.38.0/lib/overcommit/hook/pre_push/protected_branches.rb:25:in `fnmatch'
/Users/hasghari/.rvm/gems/ruby-2.2.5@ue/gems/overcommit-0.38.0/lib/overcommit/hook/pre_push/protected_branches.rb:25:in `block in protected?'
/Users/hasghari/.rvm/gems/ruby-2.2.5@ue/gems/overcommit-0.38.0/lib/overcommit/hook/pre_push/protected_branches.rb:24:in `any?'
/Users/hasghari/.rvm/gems/ruby-2.2.5@ue/gems/overcommit-0.38.0/lib/overcommit/hook/pre_push/protected_branches.rb:24:in `protected?'
/Users/hasghari/.rvm/gems/ruby-2.2.5@ue/gems/overcommit-0.38.0/lib/overcommit/hook/pre_push/protected_branches.rb:18:in `block in illegal_pushes'
/Users/hasghari/.rvm/gems/ruby-2.2.5@ue/gems/overcommit-0.38.0/lib/overcommit/hook/pre_push/protected_branches.rb:17:in `select'
/Users/hasghari/.rvm/gems/ruby-2.2.5@ue/gems/overcommit-0.38.0/lib/overcommit/hook/pre_push/protected_branches.rb:17:in `illegal_pushes'
/Users/hasghari/.rvm/gems/ruby-2.2.5@ue/gems/overcommit-0.38.0/lib/overcommit/hook/pre_push/protected_branches.rb:5:in `run'
/Users/hasghari/.rvm/gems/ruby-2.2.5@ue/gems/overcommit-0.38.0/lib/overcommit/hook/base.rb:45:in `block in run_and_transform'
/Users/hasghari/.rvm/gems/ruby-2.2.5@ue/gems/overcommit-0.38.0/lib/overcommit/utils.rb:260:in `with_environment'
[...]

PR #481, opened 2017-03-23
ProtectedBranches should be able to handle pushing tags gracefully

hasghari:
This is a fix for Issue #480 

When pushing tags, the pushed ref will look something like this:
```
refs/tags/release-0.1 e0250d9551467b32e2292ff163961b3af0f8ac67 refs/tags/release-0.1 0000000000000000000000000000000000000000
```

This does not match the regular expression pattern of `refs/heads`, hence the failure.

### DIFF
--- a/lib/overcommit/hook/pre_push/protected_branches.rb
+++ b/lib/overcommit/hook/pre_push/protected_branches.rb
@@ -21,6 +21,7 @@ module Overcommit::Hook::PrePush
 
     def protected?(remote_ref)
       ref_name = remote_ref[%r{refs/heads/(.*)}, 1]
+      return false if ref_name.nil?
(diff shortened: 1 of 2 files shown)


sds:
Thanks for the PR, @hasghari!

Would be nice if we could protect against pushing arbitrary refs, but feels like this solution satisfies the common case.

The failure on Windows seems to be spurious (I've retried that build multiple times at it seems to always fail on a different version of Ruby, so the lack of consistency tells me the failure is not caused by your changes.

hasghari:
Thanks for the merge @sds. Can we get a patch release for this fix?

sds:
Looking to release a new version soon, @hasghari. Just waiting for #479 to be wrapped up. Thanks for your patience!

Issue #483, opened 2017-04-03
haml-lint: Does not honor per-linter excludes from .haml-lint.yml


shaneog:
I am running haml-lint via overcommit.
When I set up the `.haml-lint.yml` file and run `haml-lint` manually, I have no lint offenses.
However, when I enable overcommit and/or run `overcommit --run` the  `exclude` is not honored.

If I change the `exclude` to `enabled: false` then the linter is correctly disabled.

```yaml
linters:
  ConsecutiveComments:
    exclude:
      - 'app/views/layouts/partials/_head_javascript.html.haml'

```

I am also receiving the error when overcommit `pre-commit` or `--run` is executed (haml-lint) specific:
> Unexpected output: unable to determine line number or type of error/warning for output:
> 1 file inspected, 1 lint detected
> ✗ One or more pre-commit hooks failed

Gems:
```
haml_lint (0.24.0)
overcommit (0.39.0)
```

trotzig:
Hi @shaneog, let me see if I can help you out. 

First, is this the config that you expect to apply to haml_lint? 

```
linters:
  ConsecutiveComments:
    exclude:
      - 'app/views/layouts/partials/_head_javascript.html.haml'
```

It looks like it belongs to a custom check named "ConsecutiveComments". Which could explain why the excludes config is not working. I might be missing something though.

shaneog:
Thanks @trotzig.

`ConsecutiveComments` is a linter provided [within haml-lint](https://github.com/brigade/haml-lint/blob/master/config/default.yml#L23).

I have made a minimum test case to reproduce the bug: https://github.com/shaneog/overcommit-haml-bug

There is a README file to explain everything there.

trotzig:
Ah, of course! I'm one of the maintainers of haml-lint too, so I should have noticed the config not belonging to overcommit. 

Let me have a look at that repro case. Thanks for providing a way to reproduce!

mbeasley:
If this is any help, there may be multiple issues going on. @shaneog I came across the same error that you showed in your bug repo and the problem I found was that haml-lint changed their default reporter to include the summary line: `1 file inspected, 1 lint detected` which overcommit can't parse.

I think maybe it would make sense for haml-lint to report summary info to STDERR and only the actual lines to STDOUT, **but** a work around in overcommit is to add `flags: --no-summary` to your overcommit configuration for haml-lint.

I don't know if that will have any effect on your other part of this issue, about overcommit respecting the haml-lint config.

[...]

PR #484, opened 2017-04-04
Add `--no-summary` flag to haml-lint

trotzig:
Haml-lint now prints a summary at the end of the run. E.g.

  11 files inspected, 1 lint detected

We need to disable that so that overcommit can parse the output
correctly.

Here's some more context: https://github.com/brigade/haml-lint/pull/176

This will help resolve some of the issues seen in #483.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -288,6 +288,7 @@ PreCommit:
     description: 'Analyze with haml-lint'
     required_executable: 'haml-lint'
     install_command: 'gem install haml-lint'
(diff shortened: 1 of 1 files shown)


trotzig:
Ping @michaelherold, thanks for providing that flag in haml-lint. :)

trotzig:
Ah, good point. I hadn't thought of that, but I don't see any other way than to just introduce it and force people to upgrade. 

I'm ignoring the appveyor tests for now.

PR #488, opened 2017-04-23
Add Pronto pre-commit hook

mmozuras:
👋

PR #489, opened 2017-04-23
Fix Pronto pre-commit hook flags configuration

mmozuras:
@sds well, this is embarrassing 😄. It took me an hour or two to discover that I've made a bug with my [initial PR](https://github.com/brigade/overcommit/pull/488). I've tested on a branch that had warnings without `--staged` option. As I've started actually using on real commits/branches, I quickly recognised that something is wrong. 😅


sds:
Thanks for catching and fixing!

PR #493, opened 2017-05-31
Rename Pep8 and Pep257 Hooks

proinsias:
The `pep8` and `pep257` projects have been renamed to `pycodestyle`
and `pydocstyle`, respectively.

To reflect these changes:

* The references to the former projects in `README.md` and `default.yml`
  have been updated.

* The ruby files under `pre_commit` for the former projects have been
  kept for backwards compatibility with existing `.overcommit.yml`
  files.

* New ruby files have been created under `pre-commit`.


proinsias:
@trotzig – so that an existing `.overcommit.yml` with a `Pep8` section would result in `overcommit` calling `pycodestyle`? I was assuming that the user might not have `pycodestyle` installed yet, but would likely still have `pep8` installed.

trotzig:
I was thinking we could just inject the name of the binary to some shared code. But yeah, probably not worth the effort. 

One thing we could do is to add a comment in the default config at the place for the old hooks to mention that they are deprecated.

trotzig:
The appveyor failure looks unrelated to your change. @sds, looks like it could be failing because of changes you made to `LineEndings` in afdebb. Do you have any ideas?

proinsias:
@trotzig – I've added comments to the default config as requested.

trotzig:
Thanks!

sds:
@trotzig I don't have access to a Windows machine to look into that failure, but given the test that's failing is for a pretty minor feature, I've changed it to not run on Windows environments for now (see  d44631403a3c06b08ffbe818eef58cb4cf7e91e6).

PR #496, opened 2017-06-21
Fix support for binary files in LineEndings hook

ajgon:
`LineEndings` hook checks `git ls-files --eol` indexes for `l/-text` for some reason. However, the proper format (at least for git 2.11 and up is `i/-text` (`i` instead of `l`).

This fix corrects that, while keeping backwards compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Overcommit Changelog
 
+## master
+
+* Fix binary files detection in `LineEndings` pre-commit hook
+
 ## 0.40.0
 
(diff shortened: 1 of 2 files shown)


trotzig:
Thanks @ajgon! I'm trying to figure out why the specs fail in windows environments. Are you able to explore this? I'm swamped at the moment. Perhaps push a revert of your commit and see if it still breaks?

ajgon:
Tried some configurations, and couldn't determine why it behaves this way. I'm closing it, since apparently, the problem occurs only for me, so there is no point of introducing a possible regression.

PR #497, opened 2017-06-28
Re-add hard tab exclusions for makefiles and Go

sectioneight:
This partially reverts 3f7c037b after discussion with @sds.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -304,6 +304,9 @@ PreCommit:
     quiet: true
     required_executable: 'grep'
     flags: ['-IHn', "\t"]
+    exclude:
+      - '**/Makefile'
+      - '**/*.go'
 
   Hlint:
     enabled: false


sectioneight:
NOTE: I haven't added tests **yet** as I wanted to get feedback on the approach. I'm surprised I haven't run into this before since I've been using overcommit and Make/Go for so long... am I missing something obvious?

lib/overcommit/hook/pre_commit/hard_tabs.rb
```diff
@@ -9,5 +9,14 @@ def run
         /^(?<file>(?:\w:)?[^:]+):(?<line>\d+)/,
       )
     end
+
+    def applicable_files
+      # Filter out `Makefile`, `makefile`, and `*.go` by default, since hard
+      # tabs are *required* in all these languages
+      super.reject do |modified|
+        fname = File.basename(modified)
+        fname =~ /Makefile/i or fname.end_with? '.go'
```
trotzig: Would it make sense to be a little more strict on the makefile regexp? So that we don't catch things like `makeFileArray.js` etc.

sds:
Any reason we couldn't use the `exclude` option on the hook configuration itself to accomplish this?

```yaml
  HardTabs:
    exclude: 
      - '**/Makefile'
      - '**/Makefile.*'
      - '**/makefile.*'
      - '**/*.go'
```

Unfortunately Ruby globs don't support alternation, so there's some duplication here if you want case-insensitivity.

As for why having to add this is a surprise, this is probably because I removed some default exclusions in 3f7c037b960e4a933ed4d22d97a00ac9010afd6b. My reasoning at the time was that you already had to opt-in to use the hook, so it was up to you to also specify which files it considered. My other reasoning was that I was expecting to add better support for sharing configurations across multiple projects, but this never happened. 

I admit I'm more open to setting defaults here, since it's overwhelmingly the case that `.go` will refer to Golang files. Feel free to update the PR appropriately; no need to add tests.

sectioneight:
@sds it seems silly to have to have all users add the same `exclude` blocks when we **know** that there are going to be hard-tabs in .go and Makefiles. Unless there's some other way for me to express that this hook has a **default** set of excludes. That's why I was putting this up as a request for comment, not necessarily a "real" diff.

lib/overcommit/hook/pre_commit/hard_tabs.rb
sectioneight: Yea, good catch

sds:
Right, I was trying to suggest to add those exclusions to the default `HardTabs` configuration in https://github.com/brigade/overcommit/blob/d44631403a3c06b08ffbe818eef58cb4cf7e91e6/config/default.yml#L301-L306

This would apply to all users of Overcommit going forward.

I'm on your side here. I was simply trying to convey why I originally removed the default. My fear was that Overcommit would start receiving pull requests for file extensions of questionable importance, and I didn't want to have to start investigating the "worthiness" of these extensions and whether they warranted being a default.

My hope was to have Overcommit support reusable configurations so that, for example, you could include an "overcommit-golang" gem (or something like that) which would immediately configure Overcommit with defaults appropriate for Golang projects. I never got around to this.

At the end of the day, Makefiles and Golang files are popular enough that setting them as defaults makes sense, so I'm all for adding them to the default `exclude` for `HardTabs`. Does that make sense?

sectioneight:
Makes total sense. I was missing "default.yml". Will update the PR.

sds:
Released [0.40.0](https://github.com/brigade/overcommit/releases/tag/v0.40.0). Hope that helps!

sectioneight:
\o/ thanks

PR #498, opened 2017-07-11
Allow hooks to be provided in separate gems

saverio-kantox:
When you have several repositories, and a set of common hooks, it is very complex to update all the hooks in each repository. With this change, "built in" hooks are searched in the whole load path, so the user can put one or more hooks in a gem and manage them consistently

### DIFF
--- a/lib/overcommit/configuration.rb
+++ b/lib/overcommit/configuration.rb
@@ -1,5 +1,6 @@
 require 'digest'
 require 'json'
+require 'English'
 
 module Overcommit
   # Stores configuration for Overcommit and the hooks it runs.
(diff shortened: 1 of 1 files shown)


lib/overcommit/configuration.rb
```diff
@@ -1,5 +1,6 @@
 require 'digest'
 require 'json'
+require 'English'
```
sds: Are you sure this is necessary? I'm able to access `$LOAD_PATH` in an IRB session without `require 'English'`.

lib/overcommit/configuration.rb
```diff
@@ -264,8 +265,10 @@ def ad_hoc_hook?(hook_context, hook_name)
     def built_in_hook?(hook_context, hook_name)
       hook_name = Overcommit::Utils.snake_case(hook_name)
 
-      File.exist?(File.join(Overcommit::HOME, 'lib', 'overcommit', 'hook',
-                            hook_context.hook_type_name, "#{hook_name}.rb"))
+      $LOAD_PATH.any? do |dir|
```
sds: I'm wary of this approach because it bypasses the standard mechanism for loading hook plugins, which ensures the source code of the plugins is properly signed.

By allowing code loading this way, an attacker can easily inject malicious code by simply adding a gem to the Gemfile specified in the `.overcommit.yml` `gemfile` option. We need to have a more explicit way of specifying which code is loaded so that Overcommit can properly sign the hooks to protect against this attack.

sds:
As mentioned in previous comments, we're definitely open to this feature, but we'd like an implementation that doesn't bypass the security provided by hook signing. 

Feel free to open another PR. Thanks!

lib/overcommit/configuration.rb
saverio-kantox: You probably have `English` loaded by IRB itself. I have been bitten by this previously, so now I require `English` whenever I use english global constants.

PR #502, opened 2017-07-25
Adding ruby 2.4.1 to travis-ci config

Br3nda:
I'm using overcommit on ruby 2.4.1 .. it works sweet as!. Can we add it to the supported list by including in travis builds?


sds:
Awesome, thanks @Br3nda!

Issue #300, opened 2015-11-26
Can't rebase due to signature verification


deivid-rodriguez:
I can't rebase to past revisions with different overcommit configurations. I end up in a detached state. I tried disabling signature verification but it will only work for commits _after_ disabling the check.

This behavior is weird since I've already acknowledged all past revisions of overcommit's config, so I shouldn't need to verify them again. The only solution I've found it to completely uninstall overcommit hooks just for the rebase.

Let me know if you need more information.

Thanks!

trotzig:
Echoing what @sds informed me about when I ran into the same issue: 

> Because of the nature of the latest overcommit version, you needed to pull/rebase with `OVERCOMMIT_DISABLE=1`. 

So something like
```bash
OVERCOMMIT_DISABLE=1 git pull origin master
```

deivid-rodriguez:
@trotzig @sds Thanks to both for that. It's certainly more convenient than the workaround I ended up using:

```
bundle exec overcommit --uninstall
git rebase ...
bundle exec overcommit --install
```

I still think this is undesirable and shouldn't be necessary but if the fix is difficult to implement, I find the workaround acceptable.

sds:
You raise a good point, @deivid-rodriguez. It would be nice if Overcommit remembered all previously-signed configurations.

gerrywastaken:
@sds also even if it hasn't remembered previous states, it probably shouldn't end up leaving the branch in a detached state with no obvious way to recover.

ryanclark2:
Is this still alive? Ran into this today after making changes to `overcommit.yml` then rebasing a branch onto master.

benbernard:
would really like this to land.  Assume you folks would accept a pull request? Any pointers for where to start with the signatures?

sds:
Would _love_ a PR addressing this, @benbernard! 

You'll need to modify the [`HookSigner`](https://github.com/brigade/overcommit/blob/v0.40.0/lib/overcommit/hook_signer.rb) class to support searching through a list of previously-verified signatures. 

This may involve creating a different storage format for the configuration value so that a quick lookup can be performed (emphasis on quick&mdash;the list of signatures will potentially grow quite large with time, so this needs to be something supporting a constant-time lookup). 

Hope that helps!

benbernard:
[...]

PR #505, opened 2017-08-10
Adding Signature History Support

benbernard:
This is in reference to #300 

I am very inexperienced with ruby, so I've probably done some strange / wrong things here, would love your feedback on anything like that.

I also haven't written tests yet.  I wanted to get this up here and gather any feedback on the approach, and I have a couple of questions

1. Should I remove the .git/config based stuff all together? I considered doing this, but decided not to in the thought that keeping it in place meant that I could basically guarantee that we weren't affecting performance very much.
2. I couldn't find any tests for the configuration stuff, which seems to have been duplicated from hook_signer (well one was duplicated from the other at least).  Did I just miss those? I assume I should write some if they don't already exist
3. Feedback on writing tests? I was thinking: Check that if you sign a new signature and then return to an old signature, that it doesn't think the configuration has changed.  I wasn't necessarily going to peak into the files for the tests, but rather just test the class interface.
4. What is your commit / merging style? I normally commit a lot, but am happy to squash down to one commit if that is what you folks want

Obviously, not ready to merge, and just looking for feedback here.


lib/overcommit/configuration.rb
```diff
@@ -37,6 +39,22 @@ def plugin_directory
       File.join(Overcommit::Utils.repo_root, @hash['plugin_directory'] || '.git-hooks')
     end
 
+    # Returns absolute path to directoy for the history of signatures
```
sds: directoy -> directory

lib/overcommit/configuration.rb
```diff
@@ -37,6 +39,22 @@ def plugin_directory
...
+    # @return [String] - Absolute path to the signature directory
+    def signature_directory
+      File.join(Overcommit::Utils.repo_root,
+                '.git',
+                @hash['signature_directory'] || 'overcommit-signatures')
+    end
+
+    def signature_history
+      if @hash['signature_directory']
+        @hash['signature_history'].to_i
```
sds: Can you elaborate on why setting the signature history is dependent on whether the `signature_directory` configuration option is set? It seems like we should respect `signature_history` regardless.

lib/overcommit/signer.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    private
+
+    def add_signature_to_history(signature)
+      # Now we must update the history file with the new signature
+      # We want to put the newest signatures at the top, since they are more
+      # likely to be used, and will be read sooner
+      signatures = []
+      if has_history_file
+        signatures = (File.readlines history_file).first(@config.signature_history - 1)
```
sds: The parentheses here seem odd. Let's just do:

```ruby
File.readlines(history_file).first(@config.signature_history - 1)
```

lib/overcommit/signer.rb
```diff
@@ -0,0 +1,124 @@
...
+
+      File.open(history_file, 'w') do |fh|
+        fh.write("#{signature}\n")
+        signatures.each do |old_signature|
+          fh.write(old_signature)
+        end
+      end
+    end
+
+    def signature_in_history_file(signature)
```
sds: Let's end this with `?` for consistency with other boolean-returning methods.

Same with `has_history_file` below.

lib/overcommit/signer.rb
```diff
@@ -0,0 +1,124 @@
...
+      unless has_history_file
+        return false
+      end
+
+      found = false
+      File.open(history_file, 'r') do |fh|
+        # Process the header
```
sds: What is this comment referring to?

lib/overcommit/signer.rb
```diff
@@ -0,0 +1,124 @@
...
+        until (line = fh.gets).nil?
+          line.chomp
+
+          if line == signature
+            found = true
```
sds: Could simply `return true` here and then `false` at the end of this method since there doesn't seem to be a strong need for the `found` variable.

lib/overcommit/configuration.rb
```diff
@@ -37,6 +39,22 @@ def plugin_directory
       File.join(Overcommit::Utils.repo_root, @hash['plugin_directory'] || '.git-hooks')
     end
 
+    # Returns absolute path to directoy for the history of signatures
+    # @return [String] - Absolute path to the signature directory
+    def signature_directory
+      File.join(Overcommit::Utils.repo_root,
+                '.git',
+                @hash['signature_directory'] || 'overcommit-signatures')
```
sds: Can you elaborate on why allowing this to be customized is useful? I'm fearful of an overzealous user changing this to a directory of files which are tracked by their Git repository because it can be "helpful" to others in storing previous signatures. Of course this is problematic since now an attacker can modify said files.

I anticipate this is a highly-unlikely scenario, but I would prefer to stick to the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle unless there's a solid reason to allow this to be customized. Curious to hear your thoughts here.

lib/overcommit/signer.rb
sds: Ah, I see you might have done this because of the block, perhaps for readability. Keep it if you like.

proinsias:
Any change of an update here? I'd love to see this functionality.

benbernard:
Haven't had a chance to get back to this, all the comments make sense, just haven't had a chance to make tests

proinsias:
Totally understand – just showing my interest! Thanks for working on this!

benbernard:
I'm not using overcommit anymore (haven't in years, i'm going to close this PR, if anyone wants to take this over the line, feel free

Issue #292, opened 2015-11-07
Possible contribution of PHP tools


craig-davis:
I have initial support for PHP linting via `php -l` and PHP code style checks via [phpcs](https://github.com/squizlabs/PHP_CodeSniffer). In the future, I plan to implement other tools from the [PHP QA Toolchain](http://phpqatools.org/).

Would you like a pull request for these initial hooks?

lencioni:
Yes, absolutely we would. Please be sure to include specs and thanks for offering to help out!

craig-davis:
Hi @lencioni - Where do you prefer that fixture files be located? I was expecting `spec/overcommit/fixtures` but don't see anything there.

PR #507, opened 2017-08-17
Add initial support for phpcs and php lint

craig-davis:
This addresses #292 to add PHP commit hooks - a much belated contribution!

This PR adds two new linters:
  * PHP Lint via the `php -l` command to check for valid php syntax and confirms that the file will compile
  * PHP Code Sniffer via `phpcs` to check for style problems

This can later be followed by other php tooling, but these are the most common and I believe are a good starting point.

This may not be entirely idiomatic Ruby, please feel free to critique and suggest improvements. I've added a collection of specs that should cover the most common cases.


craig-davis:
It looks like this PR has some line ending issues that didn't turn up during local testing. I'll investigate this and update the PR as soon as possible.

sds:
Thanks for the contribution, @craig-davis!

craig-davis:
@sds Thank you!

We use Overcommit at my employer [Help Scout](https://www.helpscout.net/) in many of our projects and it's been a valuable tool. I hope this contribution will be the start of adding some other PHP tools by the community.

PR #509, opened 2017-08-25
Optionally disable colorized output via env

### LINKED ISSUE #508: Optionally disable colorized output
Would it be possible to always disable colorization of log output?
I'm running into the issue described here: https://groups.google.com/forum/#!topic/magit/oMNrMUvxTTs

It seems like `Overcommit::Logger#color` could additionally read some configuration option (environment variable?) and decide whether to include the ANSI escape sequences or not.

Would you accept a PR along those lines?

harto:
Colorized output is now disabled if the environment variable `OVERCOMMIT_COLOR` is present with the value `0`.

Fixes #508.

### DIFF
--- a/lib/overcommit/logger.rb
+++ b/lib/overcommit/logger.rb
@@ -78,7 +78,11 @@ module Overcommit
     # @param partial [true,false] whether to omit a newline
     def color(code, str, partial = false)
       send(partial ? :partial : :log,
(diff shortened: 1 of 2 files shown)


sds:
Thanks for the pull request, @harto!

harto:
Thank you! :)

Issue #470, opened 2017-02-13
Add support for prepare-commit-msg hooks


siwica:
I want to modify the default commit message a user sees when opening the commit message editor. Specifically I want to replace some placeholder in the commit-template with a string extracted from the current branch name and display the resulting text to the user without modifying the original commit template file.

Can this be done using overcommit and if so how and using which hook? `pre_commit` or `prepare_commit_msg`?

trotzig:
Would this help? https://robots.thoughtbot.com/better-commit-messages-with-a-gitmessage-template

sds:
Hey @siwica, 

You are right that a [`prepare-commit-msg`](https://git-scm.com/docs/githooks#_prepare_commit_msg) hook is the correct way to go here since you want to pre-populate a commit message with dynamic content _before_ allowing the user to edit the commit message.

Overcommit does not yet support `prepare-commit-msg` hooks, but we'd happily accept a pull request adding support!

siwica:
I might take a look at it, if I find some time. Haven't done anything in Ruby for quite a while though.

Could you shortly outline what would need to be done in order to add support for `prepare-commit-msg`  hooks?

sds:
You can see what went into adding support for `pre-rebase` hooks in #165. 

According to the [`prepare-commit-msg`](https://git-scm.com/docs/githooks#_prepare_commit_msg) documentation, it will need to support one to three arguments, so you'll want to expose helpers for accessing these arguments in the `Overcommit::HookContext::PrepareCommitMsg` class you will create.

The `Overcommit::HookContext::PrepareCommitMsg` class you create will need to provide a way to edit the commit message, similar to the [`Overcommit::HookContext::CommitMsg` class](https://github.com/brigade/overcommit/blob/d486e96d93dd58a7f96a22a0365cc11de8164098/lib/overcommit/hook_context/commit_msg.rb#L14-L18).

siwica:
Alright, thank you!

-  What should a helper return when an argument can only be one out of a finite number of options (in my case: `message`, `template`, `squash` or `commit`). Return the raw string `'message''`, or rather a symbol `:message`?
- How should a helper behave if an argument is not set (`prepare-commit-msg` takes a variable number of arguments) and is thus not available? Raise an error or return `nil`?

sds:
Sorry for the delay in response, @siwica.

[...]

PR #516, opened 2017-10-20
MessageTemplate post-checkout hook

dcluna:
This commit adds the MessageTemplate post-checkout hook, which, when correctly
configured, automatically generates a [commit message
template](https://robots.thoughtbot.com/better-commit-messages-with-a-gitmessage-template)
based on the branch name.

Groups captured in the `branch_pattern` regex can be used in `replacement_text`;
see the accompanying spec for details. Also, `replacement_text` can be a path to
a file, whose text will be processed following the same rules.

--

This may help with the request in #470.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -785,6 +785,14 @@ PostCheckout:
     quiet: true
     recursive: false
 
+  MessageTemplate:
+    enabled: false
+    description: 'Write commit template message based on the new branch name'
+    quiet: true
+    recursive: false
+    branch_pattern: '\A.*\w+[-_](\d+).*\z'
+    replacement_text: '[#\1] [ci skip]'
+
 # Hooks that run after a commit is created.
 PostCommit:
   ALL:
--- a/lib/overcommit/hook/post_checkout/message_template.rb
+++ b/lib/overcommit/hook/post_checkout/message_template.rb
@@ -1,0 +1,50 @@
+module Overcommit::Hook::PostCheckout
+  class MessageTemplate < Base
+    def run
+      Overcommit::Utils.log.debug("Checking if '#{Overcommit::GitRepo.current_branch}' matches #{branch_pattern}")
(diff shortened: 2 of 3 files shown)


sds:
Hey @dcluna, thanks for sharing this PR. Apologies for not responding to your [original request for feedback in #470](https://github.com/brigade/overcommit/issues/470#issuecomment-330064036).

Let me first say that I totally see the value in what you are trying to accomplish. Having the ability to customize the commit message template that appears when you commit is a useful feature, especially in organizations that value well-written commit messages.

However, I disagree with the approach taken here to accomplish the goal of customizable commit message templates. 

First of all, the `prepare-commit-msg` hook is explicitly meant for this use case, so it seems odd to piggy-back on the `post-checkout` hook. Furthermore, using a `post-checkout` hook results in this code running **every** time your `HEAD` changes, rather than just at the time you are creating your commit. Lastly, a `prepare-commit-msg` hook would give more flexibility that what you currently have here, and be more extensible (for example, you could inspect the staged files and generate a template depending on which files were changed, which you can't do with a `post-checkout` since you haven't authored anything at that point).

What this implementation does is perfectly suitable for a single organization to use if it chooses, but I don't see it as extensible or general-purpose enough to be included by default in Overcommit.

The better long-term answer is to implement #470 so that arbitrarily-complex templates can be generated on an as-needed basis, rather than every time `HEAD` changes. This uses the existing tools provided by Git as intended, and also provides an entire platform for developers to build on top of, rather than a less-flexible solution limited to branch names and regex substitutions.

Hope that explains my position here. Would be more than happy to accept a pull request implementing #470! (which is not that much work&mdash;see [my comment](https://github.com/brigade/overcommit/issues/470#issuecomment-279920224) in the original thread)

dcluna:
@sds Thanks for the kind words. If nobody beats me to it, I'll try to come up with the PR you suggested.

PR #517, opened 2017-10-24
Add support for Puppet metadata-json-lint pre-commit hook

hajee:


lib/overcommit/hook/pre_commit/metadata_json_lint.rb
```diff
@@ -0,0 +1,50 @@
...
+          "--#{name}"
+        else
+          "--no-#{name}"
+        end
+      end
+    end
+
+    def run
+      # When metadata.json, not modified return pass
+      return :pass unless applicable_files.include?('metadata.json')
```
sds: Rather than checking for the `metadata.json` file here (which would mean you run this hook _every_ `pre-commit`, which is inefficient), you should configure the hook itself (see my comment on `config/default.yml`).

README.md
```diff
@@ -519,6 +519,7 @@ issue](https://github.com/brigade/overcommit/issues/238) for more details.
 * [LocalPathsInGemfile](lib/overcommit/hook/pre_commit/local_paths_in_gemfile.rb)
 * [Mdl](lib/overcommit/hook/pre_commit/mdl.rb)
 * [`*`MergeConflicts](lib/overcommit/hook/pre_commit/merge_conflicts.rb)
+* [MetadataJsonLint](lib/overcommit/hook/pre_commit/metadata_json_lint.rb)
```
sds: Since this tool is specific to the Puppet ecosystem, let's name it `PuppetMetadataJsonLint` to reflect this.

config/default.yml
```diff
@@ -427,6 +427,15 @@ PreCommit:
...
 
+  MetadataJsonLint:
+    enabled: false
+    strict_license: true
+    strict_dependencies: false
+    fail_on_warning: true
+    description: 'Checking Puppet module metadata'
+    required_executable: 'metadata-json-lint'
+    install_command: 'gem install metadata-json-lint semantic_puppet'
+
```
sds: Make sure this runs only when `metadata.json` changes by adding:

```yaml
  PuppetMetadataJsonLint:
    ...
    include: 'metadata.json'
```

config/default.yml
```diff
@@ -427,6 +427,15 @@ PreCommit:
...
     flags: ['-IHn', "^<<<<<<<[ \t]"]
 
+  MetadataJsonLint:
+    enabled: false
+    strict_license: true
+    strict_dependencies: false
+    fail_on_warning: true
+    description: 'Checking Puppet module metadata'
+    required_executable: 'metadata-json-lint'
+    install_command: 'gem install metadata-json-lint semantic_puppet'
```
sds: Would suggest we remove `semantic_puppet` from this list, as the `post_install` message for `metadata-json-lint` explicitly mentions installing it if you're on an older version of Puppet: https://github.com/voxpupuli/metadata-json-lint/blob/a57b7d2f21d746d616ef6032624df1a7cc183425/metadata-json-lint.gemspec#L26-L30

lib/overcommit/hook/pre_commit/metadata_json_lint.rb
```diff
@@ -0,0 +1,50 @@
...
+      [:strict_license, :strict_dependencies, :fail_on_warning].collect do |option|
+        name = option.to_s.tr('_', '-')
+        value = config.fetch(option.to_s) { true }
+        if value
+          "--#{name}"
+        else
+          "--no-#{name}"
+        end
+      end
+    end
```
sds: I would recommend against making these configurable options on the hook itself since flag names can change and you can simply add/remove flags using the `flags` option for the hook.

So for your configuration, I would change the YAML to say:

```yaml
  PuppetMetadataJsonLint:
    enabled: false
    description: 'Checking module metadata'
    flags: ['--no-strict_license', '--no-strict-dependencies', '--fail-on-warning']
    include: 'metadata.json'
```

This reduces the surface area you need to support, so if the underlying tool changes you don't need to make any changes to Overcommit to start using them.

This has the added benefit of simplifying your implementation to not require this `options` method at all:

```ruby
def run 
  result = execute(command, args: applicable_files)
  output = result.stdout.chomp.gsub(/^"|"$/, '')
  return :pass if result.success? && output.empty?

  extract_messages(
    output.split("\n"),
    MESSAGE_REGEX,
    MESSAGE_TYPE_CATEGORIZER
  )
end
```

Since you specified `include` in the configuration, `applicable_files` will always just include the one `metadata.json` file.

spec/overcommit/hook/pre_commit/metadata_json_lint_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+          result.stub(:stdout).and_return(normalize_indent(<<-OUT))
+            (ERR) requirements: The 'pe' requirement is no longer supported by the Forge.
+          OUT
+        end
+
+        it { should fail_hook }
+      end
+    end
+  end
+end
```
sds: Thank you for taking the time to write thorough tests!

hajee:
Hi,

Thanks for the speedy, thorough and educational review. Id didn't realize the framework already contained such rich functionality. I guess I should have read a bit more and do less :-).

I will apply the requested changes

hajee:
Don't know why the Windows tests fail :-(

config/default.yml
```diff
@@ -427,6 +427,14 @@ PreCommit:
     required_executable: 'grep'
     flags: ['-IHn', "^<<<<<<<[ \t]"]
 
+  PuppetMetadataJsonLint:
+    enabled: false
+    description: 'Checking module metadata'
+    flags: ['--strict_license', '--strict-dependencies', '--fail-on-warning']
```
sds: Are you sure `strict_license` is an underscore and not a hyphen?

lib/overcommit/hook/pre_commit/puppet_metadata_json_lint.rb
```diff
@@ -0,0 +1,32 @@
+module Overcommit::Hook::PreCommit
+  #
+  # Run's the Puppet metadata linter. It has support for adding options
+  # in the .overcommit.yaml
+  #
+  # PreCommit:
+  #   PuppetMetadataJsonLint:
+  #     enabled: true
+  #    flags: ['--no-strict_license', '--no-strict-dependencies', '--fail-on-warning']
```
sds: Indentation is a bit off. Also underscore should probably be hyphen.

Taking a step back, I don't see a strong need to include this example YAML as a comment since it's just duplicated from `config/default.yml`.

hajee:
Fore pushed the last requested changes so the git history stays clean.

PR #518, opened 2017-10-27
Added comma for readability

filfreire:
This is very subjective and discussible but I believe it makes the sentence more readable.


sds:
Thanks for suggestion improvements to the documentation, @filfreire.

Your edit actually changes my meaning, since I intended to treat "Git hooks" as the subject here, and not refer to just the tool Git.

However, I agree this could be worded better, and have taken a stab at this in 907e5ec2. Thanks!

PR #519, opened 2017-11-03
Print commit message when commit_msg hooks fail

### LINKED ISSUE #494: Should echo commit message when commit is denied
Overcommit should echo the user provided commit message to stdout
whenever it declines a commit, so it can be copy/pasted for a new attempt.

Currently all input is lost when a validation fails, which is very frustrating
and discourages users from writing elaborate commit messages.

henrahmagix:
>So the user can copy-paste their commit message when trying to commit
again, and not lose it. Removes frustration for users who craft their
commit messages with care.
>
>The user can always get their failed message from `.git/COMMIT_EDITMSG`
(depending on the version of git, I think), but only if they copy before
running `git commit` again, which they may forget to do sometimes.

Fix #494 

I am new to ruby, gems, and this project, so please forgive any "nasties" or other mistakes! I'm happy to change or reimplement this however the maintainers would like =)

### DIFF
--- a/lib/overcommit/hook_context/commit_msg.rb
+++ b/lib/overcommit/hook_context/commit_msg.rb
@@ -31,6 +31,10 @@ module Overcommit::HookContext
       @args[0]
     end
 
+    def post_fail_message
+      "Failed commit message:\n" + commit_message_lines.join
+    end
+
     private
 
     def raw_commit_message_lines
--- a/lib/overcommit/hook_runner.rb
+++ b/lib/overcommit/hook_runner.rb
@@ -74,7 +74,13 @@ module Overcommit
 
         print_results
 
-        !(@failed || @interrupted)
+        hook_failed = @failed || @interrupted
+
+        if hook_failed && @context.respond_to?(:post_fail_message)
+          puts @context.post_fail_message
+        end
+
+        !hook_failed
       else
         @printer.nothing_to_run
         true # Run was successful
--- a/lib/overcommit/hook_runner.rb
+++ b/lib/overcommit/hook_runner.rb
@@ -76,8 +76,9 @@ module Overcommit
 
         hook_failed = @failed || @interrupted
 
-        if hook_failed && @context.respond_to?(:post_fail_message)
-          puts @context.post_fail_message
+        if hook_failed
+          message = @context.post_fail_message
+          @printer.after(message) unless message.nil?
         end
 
         !hook_failed
--- a/lib/overcommit/hook_context/base.rb
+++ b/lib/overcommit/hook_context/base.rb
@@ -102,6 +102,13 @@ module Overcommit::HookContext
       @input_lines ||= input_string.split("\n")
     end
 
+    # Returns a message to display on failure.
+    #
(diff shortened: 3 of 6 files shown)


lib/overcommit/hook_runner.rb
```diff
@@ -74,7 +74,13 @@ def run_hooks
 
         print_results
 
-        !(@failed || @interrupted)
+        hook_failed = @failed || @interrupted
+
+        if hook_failed && @context.respond_to?(:post_fail_message)
+          puts @context.post_fail_message
```
henrahmagix: The `printer` should probably be used here, right? I wasn't sure yet how to use the `printer` for this, so I wanted to commit and get a discussion going first, since this is my first ruby PR.

lib/overcommit/hook_runner.rb
```diff
@@ -74,7 +74,13 @@ def run_hooks
 
         print_results
 
-        !(@failed || @interrupted)
+        hook_failed = @failed || @interrupted
+
+        if hook_failed && @context.respond_to?(:post_fail_message)
```
henrahmagix: I thought it was best to allow such a method to not be implemented on the individual `HookContext` subclasses, but work here if it was. Right now only `HookContext::CommitMsg` implements it, to solve #494.

lib/overcommit/hook_runner.rb
henrahmagix: I couldn't find a test for `run_hooks` so I wasn't sure how to test this. Happy to add a test if I could be pointed to where?

henrahmagix:
Bump! I'd like to know if this would be accepted before I investigate how to improve/fix this PR =)

lib/overcommit/hook_runner.rb
sds: Yes, let's use the printer class (and give it a specific method for this case). Thanks!

lib/overcommit/hook_context/commit_msg.rb
```diff
@@ -31,6 +31,10 @@ def commit_message_file
       @args[0]
     end
 
+    def post_fail_message
+      "Failed commit message:\n" + commit_message_lines.join
+    end
```
sds: Let's define a default `post_fail_message` method in the base `HookContext` class so we don't need to do the `@context.respond_to?(:post_fail_message)` below. We can have a contract where returning `nil` results in nothing being printed.

henrahmagix:
@sds No worries; thanks for responding! I've pushed the changes =)

henrahmagix:
The build failures are out of my hands unfortunately. I think the Rubocop ones are errors on lots of files that haven't been changed here, but I'm also not sure if my dev environment is pulling in more rules than the build environment.

sds:
Thanks for making those changed, @henrahmagix. I've merged in 167c13c5 and a0dafc08 and made some small modifications in 0f40c0d5 to address the RuboCop warnings. Thanks again!

henrahmagix:
Thanks so much @sds! Really appreciate it - and your method name choice is way better =D

PR #520, opened 2017-11-04
prepare-commit-message hook

dcluna:
As mentioned in [#516](https://github.com/brigade/overcommit/pull/516#issuecomment-338525028), this commit implements the `prepare-commit-msg` hook, with a default hook that prepends a message based on the branch name through regex substitution.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1154,7 +1154,7 @@ PrepareCommitMsg:
     enabled: false
     description: 'Prepends the commit message with text based on the branch name'
     branch_pattern: '\A.*\w+[-_](\d+).*\z'
-    replacement_text: '[#\1] [ci skip]'
+    replacement_text: '[#\1]'
     on_fail: warn
 
 # Hooks that run during `git push`, after remote refs have been updated but
--- a/config/default.yml
+++ b/config/default.yml
@@ -1143,6 +1143,20 @@ PostRewrite:
       - 'package.json'
       - 'yarn.lock'
 
+# Hooks that run during the `prepare-commit-msg` hook.
+PrepareCommitMsg:
+  ALL:
+    requires_files: false
+    required: false
+    quiet: false
+
+  ReplaceBranch:
+    enabled: false
+    description: 'Prepends the commit message with text based on the branch name'
+    branch_pattern: '\A.*\w+[-_](\d+).*\z'
+    replacement_text: '[#\1] [ci skip]'
+    on_fail: warn
+
 # Hooks that run during `git push`, after remote refs have been updated but
 # before any objects have been transferred.
 PrePush:
--- a/lib/overcommit/hook/prepare_commit_msg/replace_branch.rb
+++ b/lib/overcommit/hook/prepare_commit_msg/replace_branch.rb
@@ -4,7 +4,7 @@ module Overcommit::Hook::PrepareCommitMsg
   # the `branch_pattern` regex.
   class ReplaceBranch < Base
     def run
-      return :pass unless commit_message_source.empty? ||
+      return :pass unless !commit_message_source ||
         commit_message_source == :commit # NOTE: avoid 'merge' and 'rebase'
       Overcommit::Utils.log.debug(
         "Checking if '#{Overcommit::GitRepo.current_branch}' matches #{branch_pattern}"
--- a/lib/overcommit/hook/prepare_commit_msg/replace_branch.rb
+++ b/lib/overcommit/hook/prepare_commit_msg/replace_branch.rb
@@ -1,0 +1,54 @@
+module Overcommit::Hook::PrepareCommitMsg
+  # Prepends the commit message with a message based on the branch name.
+  # It's possible to reference parts of the branch name through the captures in
+  # the `branch_pattern` regex.
+  class ReplaceBranch < Base
+    def run
+      return :pass unless commit_msg_source == :commit
+      Overcommit::Utils.log.debug(
+        "Checking if '#{Overcommit::GitRepo.current_branch}' matches #{branch_pattern}"
+      )
+      if branch_pattern.match(Overcommit::GitRepo.current_branch)
+        prepend_commit_message
+        :pass
+      else
+        :warn
+      end
+    end
+
+    def prepend_commit_message
+      Overcommit::Utils.log.debug("Writing #{commit_msg_filename} with #{new_template}")
+      old_contents = File.read(commit_msg_filename)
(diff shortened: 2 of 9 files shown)


lib/overcommit/hook_context/prepare_commit_msg.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    # Returns the source of the commit message, and can be: message (if a -m or
+    # -F option was given); template (if a -t option was given or the
+    # configuration option commit.template is set); merge (if the commit is a
+    # merge or a .git/MERGE_MSG file exists); squash (if a .git/SQUASH_MSG file
+    # exists); or commit, followed by a commit SHA-1 (if a -c, -C or --amend
+    # option was given)
+    def commit_msg_source
+      (@args[1] || 'commit').to_sym
```
dcluna: While testing locally, the `args` passed in a standard commit were just the filename in question (`.git/COMMIT_EDITMSG`), which wasn't expected as per the [git docs](https://git-scm.com/docs/githooks#_prepare_commit_msg). Do you have more context on this?

spec/overcommit/hook/prepare_commit_msg/replace_branch_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'spec_helper'
+
+describe Overcommit::Hook::PrepareCommitMsg::ReplaceBranch do
```
dcluna: This file lacks `repo` tests. Let me know if these are required for this PR to be merged.

spec/overcommit/hook_context/prepare_commit_msg_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    subject { context.commit_msg_source }
+
+    it { should == commit_msg_source }
+  end
+
+  describe '#commit' do
+    subject { context.commit }
+
+    context "source isn't :commit" do
+      it { should == `git rev-parse HEAD` }
```
dcluna: This is a bit ugly. Can you think of a better way to write this spec?

lib/overcommit/hook/prepare_commit_msg/replace_branch.rb
```diff
@@ -0,0 +1,54 @@
...
+        :pass
+      else
+        :warn
+      end
+    end
+
+    def prepend_commit_message
+      Overcommit::Utils.log.debug("Writing #{commit_msg_filename} with #{new_template}")
+      old_contents = File.read(commit_msg_filename)
+      File.open(commit_msg_filename, 'w') do |commit_file|
```
dcluna: Perhaps this code should be extracted into a utility function in `base`? `prepend_commit_msg` seems like a common-enough occurrence.

trotzig:
Sorry about the delay here. I don't have enough context to give this a proper review, so I'm going to ask @sds to see if he does.

lib/overcommit/hook_context/prepare_commit_msg.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    # Returns the commit's SHA-1.
+    # If commit_msg_source is :commit, it's passed through the command-line.
+    def commit
+      @args[2] || `git rev-parse HEAD`
```
sds: We need to be clear about the meaning of the value returned by this helper method. From my comment above, if the `commit` is defined then `arg[2]` will be the SHA-1 of the commit we're copying, but _not_ the commit about to be created. I think this will be confusing and we should be explicit.

Let's call this method `commit_message_source_ref`. While it's not a true ref I think the wording helps emphasize that it's referring to a commit hash ref, and emphasizes the connection to the `commit_message_source` helper above.

spec/overcommit/hook/prepare_commit_msg/replace_branch_spec.rb
sds: This is fine. As long as we have rigorous tests for the `PrepareCommitMsg::Base` class which validate the behavior of our commit message modifier helpers and the `commit_message_source` / `commit_message_source_ref` helpers, we can stub those in hook implementation tests.

lib/overcommit/hook_context/prepare_commit_msg.rb
```diff
@@ -0,0 +1,26 @@
...
+      @args[0]
+    end
+
+    # Returns the source of the commit message, and can be: message (if a -m or
+    # -F option was given); template (if a -t option was given or the
+    # configuration option commit.template is set); merge (if the commit is a
+    # merge or a .git/MERGE_MSG file exists); squash (if a .git/SQUASH_MSG file
+    # exists); or commit, followed by a commit SHA-1 (if a -c, -C or --amend
+    # option was given)
+    def commit_msg_source
```
sds: Let's call this `commit_message_source` (we aren't using the short `msg` form in any other consumable API).

spec/overcommit/hook_context/prepare_commit_msg_spec.rb
sds: You'll be able to remove this particular test given my comment above.

bpaul:
@dcluna Is this still active? I'd love to see prepare-commit-message support merged. I'm happy to tidy up this PR if necessary...

dcluna:
@bpaul sorry, I haven't had enough time to tidy this up. Feel free to move this forward

siwica:
@bpaul Hey, are you still planning on tidying this up? I'd be very much interested in having this hook as part of overcommit

config/default.yml
```diff
@@ -956,6 +956,21 @@ PostRewrite:
...
+  ALL:
+    requires_files: false
+    required: false
+    quiet: false
+
+  ReplaceBranch:
+    enabled: false
+    description: 'Prepends the commit message with text based on the branch name'
+    branch_pattern: '\A.*\w+[-_](\d+).*\z'
+    replacement_text: '[#\1] [ci skip]'
```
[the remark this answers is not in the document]
dcluna: This hook's main function is to match a ticket identifier in the beginning of the branch name, so if you follow a certain naming convention for your branches, like `$BUGTRACKER_NAME-$ID_IN_BUGTRACKER-$BRANCH_DESCRIPTION`, you'll have those already set up in the commit message name.

I use this for Github's JIRA integration, not sure if I should remove the `[ci skip]` to a hook of its own or just remove it altogether. WDYT?

sds:
Thanks @dcluna!

PR #521, opened 2017-11-13
Add stylelint to pre-commit hook

### LINKED ISSUE #514: Support for StyleLint in Pre-Commit Hook.
Hi everyone and thanks for making such a nice tool. I was recently introduced to this and I've since used it in its basic config with Rubocop and ESLint. 

There is a tool I'd love to see integrated called [StyleLint](https://github.com/stylelint/stylelint), which is used where I work, and I'd love to see it integrated into Overcommit. 

A question: If I was to attempt to build something similar to [this](https://github.com/brigade/overcommit/blob/master/lib/overcommit/hook/pre_commit/css_lint.rb) over the next few days, what are the considerations I need to take into mind?

brianbancroft:
The following is a basic implementation of StyleLint on the pre-commit hook. It's not quite ready for prime-time as I have two areas I still haven't figured out:

- Testing. I'm still stuck on how to test it
- `Unexpected output` error (see screencap 1) (regex. Still looking at it)

#### Bad CSS
Below in another repo, I added the gem locally, and added some gobbedly-gook in the CSS. This is the output:

image

#### Good CSS

Here's what it looks like otherwise:
image

Closes #514 if resolved

### DIFF
--- a/lib/overcommit/version.rb
+++ b/lib/overcommit/version.rb
@@ -2,5 +2,5 @@
 
 # Defines the gem version.
 module Overcommit
-  VERSION = '0.41.0'.freeze
+  VERSION = '0.42.0'.freeze
 end
--- a/lib/overcommit/hook/pre_commit/style_lint.rb
+++ b/lib/overcommit/hook/pre_commit/style_lint.rb
@@ -3,7 +3,10 @@ module Overcommit::Hook::PreCommit
   #
   # @see https://github.com/stylelint/stylelint
   class StyleLint < Base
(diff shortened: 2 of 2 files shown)


lib/overcommit/version.rb
```diff
@@ -2,5 +2,5 @@
 
 # Defines the gem version.
 module Overcommit
-  VERSION = '0.41.0'.freeze
+  VERSION = '0.42.0'.freeze
```
sds: Let's leave this out for now. We'll update the version when we release the next version. Thanks!

sds:
Finished in #590

PR #526, opened 2017-12-01
Fix mdl MESSAGE_REGEX

proinsias:
While reviewing the `MESSAGE_REGEX` for `mdl`, I noticed that it is incomplete, and is missing the message output of `mdl`. This updates the regular expression to specifically match this output.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -415,7 +415,7 @@ PreCommit:
 
   Mdl:
     enabled: false
-    description: 'Analyze with mdl'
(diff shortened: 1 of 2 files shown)


config/default.yml
```diff
@@ -415,7 +415,7 @@ PreCommit:
 
   Mdl:
     enabled: false
-    description: 'Analyze with mdl'
+    description: 'Analyze markdown files with mdl'
```
trotzig: 👍

Issue #529, opened 2017-12-02
GoLint pre-commit hook results in incorrect behavior


varunpatro:
Hello!

The `GoLint` pre-commit hook needs to be changed (I'm not sure of the best way) to accommodate linting over files in different packages. Let me explain further:

If I change one file `main.go`, then the `GoLint` hook runs `golint main.go` which is desirable and correct.

However, when I change files that belong to multiple packages such as:

1. `main.go`
2. `lib/util.go`

then the `GoLint` hook runs `golint main.go lib/util.go` which will always return the error (regardless of the presence of any lint errors)

> lib/util.go is in package lib, not main

This is because `golint` can only be used in the following way (from `golint --help`:

> Usage of golint:
> 	golint [flags] # runs on package in current directory
> 	golint [flags] [packages]
> 	golint [flags] [directories] # where a '/...' suffix includes all sub-directories
> 	golint [flags] [files] # all must belong to a single package

This means `golint` shouldn't be passed files that belong to different packages. 

I think a possible way to solve this is to pass `golint` the folders of the changed files, instead of the files themselves. 

I could create a PR to resolve this, but what would be the best way to mitigate this error?

sds:
Fixed by #530. Thanks for the report!

PR #530, opened 2017-12-03
Run `golint` for each file that has changed

varunpatro:
Also tracked at #529.

If I change one file `main.go`, then the `GoLint` hook runs `golint main.go` which is desirable and correct.

However, when I change files that belong to multiple packages such as:

1. `main.go`
2. `lib/util.go`

then the `GoLint` hook runs `golint main.go lib/util.go` which will always return the error (regardless of the presence of any lint errors)

> lib/util.go is in package lib, not main

This is because `golint` can only be used in the following way (from `golint --help`:

> Usage of golint:
> 	golint [flags] # runs on package in current directory
> 	golint [flags] [packages]
> 	golint [flags] [directories] # where a '/...' suffix includes all sub-directories
> 	golint [flags] [files] # all must belong to a single package

This means `golint` shouldn't be passed files that belong to different packages. 

This commit fixes this behaviour by running `golint` separately on each file that has changed. And the concatenating all their outputs.

### DIFF
--- a/lib/overcommit/hook/pre_commit/go_lint.rb
+++ b/lib/overcommit/hook/pre_commit/go_lint.rb
@@ -4,8 +4,17 @@ module Overcommit::Hook::PreCommit
   # @see https://github.com/golang/lint
   class GoLint < Base
     def run
-      result = execute(command, args: applicable_files)
-      output = result.stdout + result.stderr
+      output = ''
+      success = true
+
+      # golint doesn't accept multiple file arguments if
+      # they belong to different packages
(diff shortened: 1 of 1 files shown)


trotzig:
Thanks for the PR, and sorry for the long wait. This looks good. My only worry is performance, since now we're starting a new `golint` process for every file. We should have a plan for mitigating that if people start complaining after we merge and release this. My skills with go are very limited, is there a way to detect when something is a new package vs when it's just a subfolder? If so, we can split the command up based on package folders instead of each file.

PR #531, opened 2017-12-11
Improve performance of Gemfile.lock ignore check for BundleAudit hook

jlogsdon:
Listing ignored files in repositories with a lot of ignored files (such as build artifacts) can take a long time.


jlogsdon:
Not sure what's up with that windows failure, and I lack a windows development environment to test that against.

trotzig:
Sorry for the delay here. It looks like the windows failure is unrelated to your change. All good.

PR #532, opened 2018-01-03
Pass applicable_files argument to RailsBestPractices Hook

okeefm:
The `applicable_files` argument wasn't being passed into the `RailsBestPractices` `execute` command during the `run` method. Adding it means that `include`/`exclude` directives in the `.overcommit.yml` function properly.


trotzig:
Thanks for the PR, and sorry for the long wait. Code looks good from my end. The appveyor suite is known to be flaky, the failure looks unrelated to your change.

PR #536, opened 2018-01-14
Update Pre-Push Context Modified Files

taufek:
Previously, `Overcommit::HookContext::PrePush::modified_files` is set as `[]` which
does not give us any value in `applicable_files` as in `PreCommit` context.

This change updates the method to return list of added and updated files in current branch:

1. compared between current branch and tracking branch if tracking branch is available.

2. compared between current branch with parent branch if tracking branch is not available.

This gives us above values in `applicable_files` in `PrePush` context.

One usage example is we can create a hook to scan prohibited keywords such as `eval`, `console.log`, etc. before pushing to remote. Previously we can only do it in `PreCommit` hook but now we could also do it in `PrePush`.

I think this will also allow us to reuse existing hooks for `PreCommit` in `PrePush` context that relies on `applicable_files` method.

### DIFF
--- a/spec/overcommit/hook_context/pre_push_spec.rb
+++ b/spec/overcommit/hook_context/pre_push_spec.rb
@@ -44,6 +44,90 @@ describe Overcommit::HookContext::PrePush do
     end
   end
 
+  describe '#modified_files' do
+    subject { context.modified_files }
+
+    let(:remote_repo) do
+      repo do
+        touch 'update-me'
+        echo 'update', 'update-me'
+        touch 'delete-me'
+        echo 'delete', 'delete-me'
+        `git add . 2>&1 > #{File::NULL}`
+        `git commit -m "Initial commit" 2>&1 > #{File::NULL}`
+      end
+    end
+
+    context 'when current branch has tracking branch' do
+      let(:local_ref) { 'refs/heads/project-branch' }
+      let(:local_sha1) { get_sha1(local_ref) }
+      let(:remote_ref) { 'refs/remotes/origin/master' }
+      let(:remote_sha1) { get_sha1(remote_ref) }
+      let(:input) do
+        double('input', read: "#{local_ref} #{local_sha1} #{remote_ref} #{remote_sha1}\n")
+      end
+
+      it 'has modified files based on tracking branch' do
+        repo do
+          `git remote add origin file://#{remote_repo}`
+          `git fetch origin 2>&1 > #{File::NULL} && git reset --hard origin/master`
+
+          `git checkout -b project-branch 2>&1 > #{File::NULL}`
+          `git push -u origin project-branch 2>&1 > #{File::NULL}`
+
+          touch 'added-1'
+          echo 'add', 'added-1'
+          echo 'append', 'update-me'
+          FileUtils.rm 'delete-me'
+          `git add . 2>&1 > #{File::NULL}`
+          `git commit -m "Update Branch 1" 2>&1 > #{File::NULL}`
+
+          touch 'added-2'
+          echo 'add', 'added-2'
+          `git add . 2>&1 > #{File::NULL}`
+          `git commit -m "Update Branch 2" 2>&1 > #{File::NULL}`
+
+          should include(*%w[added-1 update-me added-2].map { |file| File.expand_path(file) })
+          should_not include(*%w[delete-me].map { |file| File.expand_path(file) })
+        end
+      end
+    end
+
+    context 'when current branch has no tracking branch' do
+      let(:local_ref) { 'refs/heads/project-branch' }
+      let(:local_sha1) { get_sha1(local_ref) }
+      let(:remote_ref) { 'refs/heads/master' }
+      let(:remote_sha1) { get_sha1(remote_ref) }
+      let(:input) do
+        double('input', read: "#{local_ref} #{local_sha1} #{remote_ref} #{remote_sha1}\n")
+      end
+
+      it 'has modified files based on parent branch' do
+        repo do
+          `git remote add origin file://#{remote_repo}`
+          `git fetch origin 2>&1 > #{File::NULL} && git reset --hard origin/master`
+
+          `git checkout -b project-branch 2>&1 > #{File::NULL}`
+
+          touch 'added-1'
+          echo 'add', 'added-1'
+          echo 'append', 'update-me'
+          FileUtils.rm 'delete-me'
+          `git add . 2>&1 > #{File::NULL}`
+          `git commit -m "Update Branch 1" 2>&1 > #{File::NULL}`
+
+          touch 'added-2'
+          echo 'add', 'added-2'
+          `git add . 2>&1 > #{File::NULL}`
+          `git commit -m "Update Branch 2" 2>&1 > #{File::NULL}`
+
+          should include(*%w[added-1 update-me added-2].map { |file| File.expand_path(file) })
+          should_not include(*%w[delete-me].map { |file| File.expand_path(file) })
+        end
+      end
+    end
+  end
+
   describe Overcommit::HookContext::PrePush::PushedRef do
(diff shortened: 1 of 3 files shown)


lib/overcommit/git_repo.rb
```diff
@@ -273,5 +273,33 @@ def branches_containing_commit(commit_ref)
...
+
+    def tracking_branch_ref
+      `git rev-parse \
+      --symbolic-full-name \
+      --abbrev-ref @{u} \
+      2> #{File::NULL}`.chomp
+    end
+
+    def parent_branch
+      `git show-branch -a \
```
sds: On a local repo of mine with a lot of remote branches, this can display warnings like:

```
warning: ignoring origin/some-remote-branch; cannot handle more than 25 refs
```

Would need to filter out stderr to avoid.

taufek: Nice catch. Will update accordingly.

taufek:
Thank you @sds for the feedback. It does make sense. I will dig further and incorporate your suggestion here.

taufek:
@sds ,

I've updated based on your feedback. What do you think?

spec/support/git_spec_helpers.rb
```diff
@@ -23,6 +23,41 @@ def repo(options = {})
...
+      end
+    end
+  end
+
+  # Retrieve sha1 based on git ref
+  #
+  # @param ref [String] git ref
+  # @return [String] ref's sha1
+  def get_sha1(ref)
+    `git show-ref --head`.
```
sds: This seems unnecessarily expensive (listing all the refs and then filtering for the one of interest). Any reason why

```bash
git rev-parse my-ref
```

...wouldn't suffice here?

spec/support/git_spec_helpers.rb
```diff
@@ -23,6 +23,41 @@ def repo(options = {})
...
+      split("\n").
+      find { |x| x =~ /#{Regexp.quote(ref)}/i }.
```
sds: Another danger here is you aren't anchoring your regex with `^` / `$`, so a ref of `foo` will match `foo`, `foobar`, `somefoobar`, etc. Fortunately we should be able to avoid entirely with `git rev-parse ...`.

spec/support/git_spec_helpers.rb
```diff
@@ -23,6 +23,41 @@ def repo(options = {})
...
+  # which points to 'some-repo' folder. The current working directory is set to
+  # the local repository's root directory.
+  def with_remote_repo
+    repo do
+      touch 'update-me'
+      `echo update >> update-me`
+      touch 'delete-me'
+      `echo delete >> delete-me`
+      `git add . 2>&1 > #{File::NULL}`
+      `git commit -m "Initial commit" 2>&1 > #{File::NULL}`
```
sds: If we intend to have `with_remote_repo` be a generic spec helper, it should create the most generic possible setup, and thus should not create any files/commits/objects.

spec/support/git_spec_helpers.rb
```diff
@@ -23,6 +23,41 @@ def repo(options = {})
...
+
+      Dir.chdir '..' do
+        remote_url = "file://#{File.expand_path('some-repo')}"
```
sds: We should not hard code `some-repo` here (that assumes we know how the `repo` helper is implemented), but obtain it from the name of the repo directory that was created.

spec/support/git_spec_helpers.rb
```diff
@@ -23,6 +23,41 @@ def repo(options = {})
...
+        `git clone "#{remote_url}" local-repo 2>&1 > #{File::NULL}`
+      end
+
+      Dir.chdir '../local-repo' do
+        yield if block_given?
```
sds: Not to be overly pedantic, but when I read `with_remote_repo` I would expect that the block I pass would execute within the context of the remote repository, but I'll admit it's a bit hard to describe what we're trying to do concisely.

spec/support/git_spec_helpers.rb
```diff
@@ -23,6 +23,41 @@ def repo(options = {})
...
+      end
+    end
+  end
```
sds: Looking at the issues with this implementation of `with_remote_repo`, I'm not entirely convinced this helper is going to get you much. We should probably stick with first principles in the specs:

```ruby
remote_repo = repo do
  # Create remote repository to our liking
  ...
end

local_repo = repo do
  `git remote add origin file://#{remote_repo}`
  `git fetch origin && git reset --hard origin/master`

  # Create local repository to our liking
  ...
  # Validate expectations
end
```

spec/overcommit/hook_context/pre_push_spec.rb
```diff
@@ -44,6 +44,75 @@
...
+
+      it 'has modified files based on tracking branch' do
+        with_remote_repo do
+          `git checkout -b project-branch 2>&1 > #{File::NULL}`
+          `git push -u origin project-branch 2>&1 > #{File::NULL}`
+
+          touch 'added-1'
+          `echo add >> added-1`
+          `echo append >> update-me`
+          `rm delete-me`
```
sds: `FileUtils.rm` (again to avoid a child process).

spec/overcommit/hook_context/pre_push_spec.rb
```diff
@@ -44,6 +44,75 @@
     end
   end
 
+  describe '#modified_files' do
+    let(:remote_url) { "file://#{File.expand_path('some-repo')}" }
```
sds: The use of `some-repo` here is cheating a bit as it assumes we know the name of the directory of the repository, which we don't (it is returned by the `repo` helper).

We can work around this by creating the remote repo in the `let` so we have the reference.

```ruby
let(:remote_repo) do
  repo do
    ...
  end
end

let(:remote_url) { "file://#{remote_repo}" }
```

You can then enter the repo in the specs to perform additional custom setup work:

```ruby
it ... do
  Dir.chdir remote_repo do
    ...
  end
end
```

taufek: Looks good. Will try this out.

spec/overcommit/hook_context/pre_push_spec.rb
```diff
@@ -44,6 +44,75 @@
...
+        double('input', read: "#{local_ref} #{local_sha1} #{remote_ref} #{remote_sha1}\n")
+      end
+
+      it 'has modified files based on tracking branch' do
+        with_remote_repo do
+          `git checkout -b project-branch 2>&1 > #{File::NULL}`
+          `git push -u origin project-branch 2>&1 > #{File::NULL}`
+
+          touch 'added-1'
+          `echo add >> added-1`
```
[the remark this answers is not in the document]
taufek: TIL. Will update accordingly.

spec/overcommit/hook_context/pre_push_spec.rb
taufek: 👍🏻

spec/support/git_spec_helpers.rb
taufek: This will work. Thanks for the tip

spec/support/git_spec_helpers.rb
taufek: My thought was to create a repo with a remote with some default files. Your comment does make sense. It could easily misinterpreted and it is not generic enough for re-use.

spec/support/git_spec_helpers.rb
taufek: Will definitely will try this out.

spec/support/git_spec_helpers.rb
taufek: 👍🏻

taufek:
@sds,

This is ready for re-review.

taufek:
Thanks @sds. Can't do it without you. I've learned so much.

PR #538, opened 2018-01-16
Add ChamberCompare hook

jfelchner:
One of the pain points that can arise from adding settings to your
application is if you've set a new (required) item in `development` and
`test`, so it works locally and the tests pass, but you've forgotten to
set it in `production`.

Then, you deploy your app only to have it blow up, which sets your
customers out with pitchforks for your head.

Fortunately Chamber provides an easy solution.  Whenever you commit
settings files, Chamber can automatically verify the settings of
multiple different namespace sets to ensure that they're all consistent.

For example, it can take `development`, `test` and `production` and
verify that all the keys that exist in one, exist in the other two.  If
that isn't the case, the hook will fail and the commit will abort.

You can enable it by creating an `.overcommit.yml` file and adding:

```yaml
ChamberCompare:
  enabled: true
```

This hook has an additional option:

| Name         | Description                                                                                                                                                                                             | Example                                       |
| ---          | ---                                                                                                                                                                                                     | ---                                           |
| `namespaces` | This is an array of arrays of namespaces to compare to each other. Typically each inner array will only have one item (the environment) but if you wanted to do a more intensive comparison, you could. | `[['development'], ['test'], ['production']]` |

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -184,13 +184,18 @@ PreCommit:
     description: 'Check for case-insensitivity conflicts'
     quiet: true
 
-  ChamberSecurity:
+  ChamberCompare:
     enabled: false
-    description: 'Check that settings have been secured with Chamber'
+    description: 'Check that settings are equivalent between namespaces'
     required_executable: 'chamber'
-    flags: ['secure', '--files']
+    flags: ['compare']
     install_command: 'gem install chamber'
-    include:
+    namespaces:
+      - ['development']
+      - ['test']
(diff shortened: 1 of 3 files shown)


jfelchner:
@sds whichever one of these hooks gets merged in first will cause a merge conflict with the other.  I don't have a problem resolving that for you.

sds:
Looks good, @jfelchner. Let me know when you've rebased #539. Thanks!

PR #540, opened 2018-01-16
Prohibited Keyword under PreCommit Hook

taufek:
Created this out of my own frustration because there was too many times I've committed things
such as `console.log` or `binding.pry` to my Pull Requests.

This is a simple check on prohibited keywords based on the list of keywords that we set in config yml.

This is useful for projects that need to blacklist few keywords/functions from being committed.


taufek:
Seems like there is already similar hook named FixMe. Closing this.

PR #543, opened 2018-01-20
Add GitLfs post-checkout, post-commit and post-merge hooks

proinsias:
#479 added the pre-push hook for `git-lfs`.

This PR adds the additional hooks that are now available with `git-lfs`.


proinsias:
Working on fixing the errors now.

lib/overcommit/hook/post_checkout/git_lfs.rb
```diff
@@ -0,0 +1,20 @@
...
+  #
+  # @see https://git-lfs.github.com/
+  class GitLfs < Base
+    def run
+      unless in_path?('git-lfs')
+        return :warn, 'This repository is configured for Git LFS but \'git-lfs\' ' \
+        "was not found on your path.\nIf you no longer wish to use Git LFS, " \
+        'disable this hook by removing or setting \'enabled: false\' for GitLFS ' \
+        'hook in your .overcommit.yml file'
+      end
```
sds: You shouldn't need to perform this check explicitly. Simply specify the `required_executable` option in the hook configuration.

I can see that you're trying to provide a bit extra information, but my hope is that the existing error message is pretty straightforward, and that anyone working with a repo using LFS knows what Git LFS is.

lib/overcommit/hook/post_checkout/git_lfs.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      result = execute(['git', 'lfs', 'post-checkout', previous_head, new_head, branch_checkout?])
```
sds: Have you read the documentation for [adding existing Git hooks](https://github.com/brigade/overcommit#adding-existing-git-hooks)? If you are just executing a ready-made hook, you don't need to write any Ruby code.

proinsias: Yes. I've had the following two issues with this approach before, and so thought it would be helpful to have specific hooks (a la the `git-lfs` `pre-push` ones) in `overcommit`:

I believe the `required_executable` has to checked into git, and so:

1. I have to put the hook manually in every repo.
2. I can't use this approach for shared repos at work, where I'm the only one using `overcommit`.

sds: Thanks for clarifying, @proinsias. I think when we added that restriction many years ago we were being overly cautious. Thinking on this, seems reasonable to allow ad hoc hooks to reference executables that are in your `PATH`. 

The danger comes from someone creating/modifying a hook that changes the command to `rm -rf /` or similar, but this should be caught by the hook signing logic for the hook's configuration itself.

I'm open to a PR that modifies the `HookSigner` to support this. Otherwise I'll get to it when I can. Thanks!

sds: Hey @proinsias, I took a stab at adding support for arbitrary commands as ad hoc hooks hooks in 5a0a4ec966a0196f395b6391a0f62324d69b6a51.

Could you try rebasing and rebuilding these hooks using that? Would love to get your feedback. Thanks!

proinsias: Looking into this now!

proinsias:
@sds – looks good to me!

PR #544, opened 2018-01-20
Extract Install Hooks Common Code

taufek:
While working on adding `ComposerInstall` hook in d9ac62e7ba9d0b8aa5c66c8998894e8e90f7d87f,
I noticed all the Install hooks mostly look the same with minor difference.

Extracted the common code to `SimpleInstall` module making things DRYer.

### DIFF
--- a/lib/overcommit/hook/shared/bower_install.rb
+++ b/lib/overcommit/hook/shared/bower_install.rb
@@ -1,13 +1,15 @@
+require 'overcommit/hook/shared/simple_install'
+
 module Overcommit::Hook::Shared
   # Shared code used by all BowerInstall hooks. Runs `bower install` when a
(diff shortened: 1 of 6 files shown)


sds:
Thanks for looking for opportunities to DRY up the code, @taufek. It's really appreciated.

In this case, I don't think the benefit is worthwhile. These implementations are already quite simple —a few lines each—and DRYing them up actually results in _more_ lines of code overall.

If I'm missing something with what you're trying to achieve, let me know, but I'm going to close this as [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) for now. Thanks!

taufek:
Understood. 👍🏻

PR #545, opened 2018-01-28
Add PhpStan (PHP Static Code Analyser) PreCommit Hook

taufek:
Added [PhpStan](https://github.com/phpstan/phpstan) PreCommit Hook.


spec/overcommit/hook/pre_commit/php_stan_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  before do
+    subject.stub(:applicable_files).and_return(%w[sample.php])
+  end
+
+  context 'when phpstan exits successfully' do
+    before do
+      sample_output = [
+        ''
+      ].join("\n")
```
sds: This might be better expressed as just an empty string.

```ruby
sample_output = ''
```

spec/overcommit/hook/pre_commit/php_stan_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      subject.stub(:execute).and_return(result)
+    end
+
+    context 'and it reports a warning' do
+      before do
+        sample_output = [
+          '/sample1.php:14:Call to an undefined static method Sample1::where()',
+          '/sample2.php:17:Anonymous function has an unused use $myVariable.'
+        ].join("\n")
+        # rubocop:enable Metrics/LineLength
```
sds: Why is this comment here? It doesn't seem to be necessary.

config/default.yml
```diff
@@ -498,6 +498,14 @@ PreCommit:
     flags: ['--standard=PSR2', '--report=csv']
     include: '**/*.php'
 
+  PhpStan:
+    description: 'Analysing PHP codes'
```
sds: Let's keep this consistent with what we do elsewhere and write:

```
Analyze with phpstan
```

lib/overcommit/hook/pre_commit/php_stan.rb
```diff
@@ -0,0 +1,30 @@
...
+  class PhpStan < Base
+    MESSAGE_REGEX = /^(?<file>.+)\:(?<line>\d+)\:(?<message>.+)/
+
+    def run
+      messages = []
+
+      applicable_files.each do |file|
+        result = execute(command, args: [file])
+        output = result.stdout.chomp
+        if result.status
```
sds: `status` is a number, which will always be truthy, which I don't think is what you intended here. Let's use

```ruby
result.success?
```

lib/overcommit/hook/pre_commit/php_stan.rb
```diff
@@ -0,0 +1,30 @@
...
+  # https://github.com/phpstan/phpstan/issues/239
+  # https://gist.github.com/edmondscommerce/89695c9cd2584fefdf540fb1c528d2c2
+  class PhpStan < Base
+    MESSAGE_REGEX = /^(?<file>.+)\:(?<line>\d+)\:(?<message>.+)/
+
+    def run
+      messages = []
+
+      applicable_files.each do |file|
+        result = execute(command, args: [file])
```
sds: Is there a reason you're only running one file at a time? This will run very slowly for a large number of files, since you're running them in serial.

Unless there's something about how the tool manages this, would strongly recommend you pass all files to `phpstan` in one run.

lib/overcommit/hook/pre_commit/php_stan.rb
```diff
@@ -0,0 +1,30 @@
...
+        output = result.stdout.chomp
```
sds: You don't need to assign this variable unless the exit status was successful, so perhaps only do this if `result.success?`

lib/overcommit/hook/pre_commit/php_stan.rb
taufek: Thanks. Will update.

lib/overcommit/hook/pre_commit/php_stan.rb
taufek: I'll give it a try.

lib/overcommit/hook/pre_commit/php_stan.rb
taufek: Will update. It does make sense.

spec/overcommit/hook/pre_commit/php_stan_spec.rb
taufek: Opps. Will remove it.

spec/overcommit/hook/pre_commit/php_stan_spec.rb
taufek: 👍

taufek:
@sds ,

Thanks for the feedback. I've updated accordingly. Ready for re-review.

PR #547, opened 2018-02-05
Modified Lines in File for PrePush Context

taufek:
Continuation from #536 where it added `modified_files` for `PrePush` context.
This is to add `modified_lines_in_file` for `PrePush` context.

### DIFF
--- a/lib/overcommit/hook_context/pre_push.rb
+++ b/lib/overcommit/hook_context/pre_push.rb
@@ -18,9 +18,17 @@ module Overcommit::HookContext
     end
 
     def modified_files
-      @modified_files ||= Overcommit::GitRepo.modified_files(
-        refs: "#{pushed_refs[0].remote_sha1}..#{pushed_refs[0].local_sha1}"
-      )
+      @modified_files ||= Overcommit::GitRepo.modified_files(refs: ref_range)
+    end
+
+    def modified_lines_in_file(file)
+      @modified_lines ||= {}
+      @modified_lines[file] =
+        Overcommit::GitRepo.extract_modified_lines(file, refs: ref_range)
+    end
+
+    def ref_range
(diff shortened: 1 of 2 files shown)


lib/overcommit/hook_context/pre_push.rb
```diff
@@ -18,9 +18,21 @@ def pushed_refs
...
+      @modified_lines ||= {}
+      unless @modified_lines[file]
+        @modified_lines[file] =
+          Overcommit::GitRepo.extract_modified_lines(file, refs: refs)
+      end
+
+      @modified_lines[file]
+    end
+
+    def refs
```
sds: Let's use the nomenclature in the [Git documentation](https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#_commit_ranges) and call this `ref_range`. I'm concerned that in this context `refs` is a bit too vague for folks looking at this code for the first time.

lib/overcommit/hook_context/pre_push.rb
```diff
@@ -18,9 +18,21 @@ def pushed_refs
...
+    end
+
+    def modified_lines_in_file(file)
+      @modified_lines ||= {}
+      unless @modified_lines[file]
+        @modified_lines[file] =
+          Overcommit::GitRepo.extract_modified_lines(file, refs: refs)
+      end
+
+      @modified_lines[file]
```
sds: You can shorten this method implementation to:

```ruby
@modified_lines ||= {}
@modified_lines[file] ||= Overcommit::GitRepo.extract_modified_lines(...)
```

taufek: Awesome.

lib/overcommit/hook_context/pre_push.rb
taufek: Agreed.

taufek:
@sds . Thanks for the feedback. This is ready for re-review.

PR #548, opened 2018-02-07
Update PrePush Modified Files and Lines with Multiple Branches Push

taufek:
In `pre-push` hook, it receives multiple ref ranges when we pushed multiple branches to remote.

Below is pushing multiple branches plus setting up the upstream branch on remote.
```
command:
git push origin -u tj-test tj-test-2

stdin:
refs/heads/tj-test 62738e20036f58060a8f94ff8cd73733acb6561b refs/heads/tj-test 0000000000000000000000000000000000000000
refs/heads/tj-test-2 feb40b515f598723fe043b7cb9510f28c287d642 refs/heads/tj-test-2 0000000000000000000000000000000000000000
```

Below is pushing multiple branches which already has upstream branch on remote.
```
command:
git push origin tj-test tj-test-2

stdin:
refs/heads/tj-test 61bc13ab2e717633441855d534904c5036f342dd refs/heads/tj-test 62738e20036f58060a8f94ff8cd73733acb6561b
refs/heads/tj-test-2 1e371d2c345e00c34410372dad2d7602aca99908 refs/heads/tj-test-2 feb40b515f598723fe043b7cb9510f28c287d642
```

With these multiple ref ranges, we should read all of them during pre-push `modified_files` and `modified_lines_in_file`.

### DIFF
--- a/lib/overcommit/hook_context/pre_push.rb
+++ b/lib/overcommit/hook_context/pre_push.rb
@@ -18,17 +18,14 @@ module Overcommit::HookContext
     end
 
     def modified_files
-      @modified_files ||= Overcommit::GitRepo.modified_files(refs: ref_range)
+      @modified_files ||= pushed_refs.map(&:modified_files).flatten.uniq
     end
 
     def modified_lines_in_file(file)
       @modified_lines ||= {}
-      @modified_lines[file] =
-        Overcommit::GitRepo.extract_modified_lines(file, refs: ref_range)
-    end
-
-    def ref_range
-      "#{pushed_refs[0].remote_sha1}..#{pushed_refs[0].local_sha1}"
+      @modified_lines[file] = pushed_refs.each_with_object(Set.new) do |pushed_ref, set|
+        set.merge(pushed_ref.modified_lines_in_file(file))
+      end
     end
 
     PushedRef = Struct.new(:local_ref, :local_sha1, :remote_ref, :remote_sha1) do
@@ -48,11 +45,23 @@ module Overcommit::HookContext
         deleted? || forced?
       end
 
+      def modified_files
+        Overcommit::GitRepo.modified_files(refs: ref_range)
+      end
+
+      def modified_lines_in_file(file)
+        Overcommit::GitRepo.extract_modified_lines(file, refs: ref_range)
+      end
+
       def to_s
         "#{local_ref} #{local_sha1} #{remote_ref} #{remote_sha1}"
       end
 
       private
+
+      def ref_range
+        "#{remote_sha1}..#{local_sha1}"
+      end
 
       def overwritten_commits
         return @overwritten_commits if defined? @overwritten_commits
--- a/spec/overcommit/hook_context/pre_push_spec.rb
+++ b/spec/overcommit/hook_context/pre_push_spec.rb
@@ -87,7 +87,55 @@ describe Overcommit::HookContext::PrePush do
           `git add . 2>&1 > #{File::NULL}`
           `git commit -m "Update Branch 2" 2>&1 > #{File::NULL}`
 
-          should include(*%w[added-1 update-me added-2].map { |file| File.expand_path(file) })
+          should == %w[added-1 added-2 update-me].map { |file| File.expand_path(file) }
+          should_not include(*%w[delete-me].map { |file| File.expand_path(file) })
+        end
+      end
+    end
+
+    context 'when pushing multiple branches at once' do
+      let(:local_ref_1) { 'refs/heads/project-branch-1' }
+      let(:local_sha1_1) { get_sha1(local_ref_1) }
+      let(:local_ref_2) { 'refs/heads/project-branch-2' }
+      let(:local_sha1_2) { get_sha1(local_ref_2) }
+      let(:remote_ref) { 'refs/remotes/origin/master' }
+      let(:remote_sha1) { get_sha1(remote_ref) }
+      let(:input) do
+        double('input', read: ref_ranges)
+      end
+      let(:ref_ranges) do
+        [
+          "#{local_ref_1} #{local_sha1_1} #{remote_ref} #{remote_sha1}\n",
+          "#{local_ref_2} #{local_sha1_2} #{remote_ref} #{remote_sha1}\n"
+        ].join
+      end
+
+      it 'has modified files based on multiple tracking branches' do
+        repo do
+          `git remote add origin file://#{remote_repo}`
+          `git fetch origin 2>&1 > #{File::NULL} && git reset --hard origin/master`
+
+          `git checkout -b project-branch-1 2>&1 > #{File::NULL}`
+          `git push -u origin project-branch-1 2>&1 > #{File::NULL}`
+
+          touch 'added-1'
+          echo 'add', 'added-1'
+          echo 'append', 'update-me'
+          FileUtils.rm 'delete-me'
+          `git add . 2>&1 > #{File::NULL}`
+          `git commit -m "Update Branch 1" 2>&1 > #{File::NULL}`
+
+          `git checkout master 2>&1 > #{File::NULL}`
+          `git checkout -b project-branch-2 2>&1 > #{File::NULL}`
+          `git push -u origin project-branch-2 2>&1 > #{File::NULL}`
+
+          echo 'append', 'update-me'
+          touch 'added-2'
+          echo 'add', 'added-2'
+          `git add . 2>&1 > #{File::NULL}`
+          `git commit -m "Update Branch 2" 2>&1 > #{File::NULL}`
+
+          should == %w[added-1 update-me added-2].map { |file| File.expand_path(file) }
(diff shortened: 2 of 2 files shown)


lib/overcommit/hook_context/pre_push.rb
```diff
@@ -18,17 +18,14 @@ def pushed_refs
...
       @modified_lines ||= {}
-      @modified_lines[file] =
-        Overcommit::GitRepo.extract_modified_lines(file, refs: ref_range)
-    end
-
-    def ref_range
-      "#{pushed_refs[0].remote_sha1}..#{pushed_refs[0].local_sha1}"
+      @modified_lines[file] = pushed_refs.map do |pushed_ref|
+        pushed_ref.modified_lines_in_file(file)
+      end.to_set.flatten
```
sds: Can you help me understand why you are using `to_set.flatten` here?

`modified_lines_in_file` returns a `Set`, which you map over to produce an array of sets that you then call `to_set` on, followed by `flatten`. This can be potentially expensive for a large number of files and/or modifies lines. 

Consider the following benchmark:

```ruby
require 'benchmark'
require 'set'

iterations = 10
sets = 100.times.map do
  Array.new(1_000) { rand(0..1_000) }.to_set
end

Benchmark.bmbm do |x|
  x.report('.to_set.flatten') do
    iterations.times do
      sets.to_set.flatten
    end
  end

  x.report('each_with_object{ merge }') do
    iterations.times do
      sets.each_with_object(Set.new) { |item, set| set.merge(item) }
    end
  end
end

puts 'Are they equivalent?'
puts sets.to_set.flatten == sets.each_with_object(Set.new) { |item, set| set.merge(item) }
```

```
Rehearsal -------------------------------------------------------------
.to_set.flatten             0.340000   0.000000   0.340000 (  0.344824)
each_with_object{ merge }   0.020000   0.000000   0.020000 (  0.016745)
---------------------------------------------------- total: 0.360000sec

                                user     system      total        real
.to_set.flatten             0.360000   0.000000   0.360000 (  0.361935)
each_with_object{ merge }   0.020000   0.000000   0.020000 (  0.016793)
Are they equivalent?
true
```

In a nutshell, it would be significantly faster for large sets of files/lines if we didn't use `to_set.flatten`, but I want to make sure I understand your reasoning here in case I missed something.

lib/overcommit/hook_context/pre_push.rb
```diff
@@ -18,17 +18,14 @@ def pushed_refs
     end
 
     def modified_files
-      @modified_files ||= Overcommit::GitRepo.modified_files(refs: ref_range)
+      @modified_files ||= pushed_refs.map(&:modified_files).flatten
```
sds: Set my comments below for how you may want to speed this up for large ref ranges containing many large file sets.

lib/overcommit/hook_context/pre_push.rb
taufek: `flatten` is called because I assume the expected output is non-nested `Set` for `modified_lines_in_file` and non-nested `Array` for `modified_files`. 

For example, `modified_lines_in_files` with 2 sets on `PushedRef` (when I pushed 2 branches) might output `[<Set(1,2)>, <Set(3,4)>]‘. Then calling`to_set.flatten`, converts it to `<Set(1,2,3,4)>`. 

I definitely will go with your suggestion if the output is the same since it has better performance.

lib/overcommit/hook_context/pre_push.rb
```diff
@@ -18,17 +18,14 @@ def pushed_refs
     end
 
     def modified_files
```
taufek: @sds ,

I want to ask your opinion on this one. If I push changes on the same file on separate branch/local_ref:

1) Should we convert the output to unique list of files.  

Or

2) Should we distinguish the file from different local_ref by appending sort of prefix. For example:

```
[‘<local_ref_1>:file_one’, ‘<local_ref_2>:file_one’]
```

Or

3) Do nothing. Let there be duplicate of file from different local_ref.

sds: Good question, @taufek. If I try to put myself in the shoes of a developer writing a pre-push hook, the least-surprising behavior to me would be net set of differences between the local and remote branches, and thus the unique list of files (option 1) seems reasonable to me.

If you have other ideas about why others would be better, let me know!

taufek: Yup. Unique list of files make sense to me. Thanks.

taufek:
@sds ,

It's ready for re-review.

lib/overcommit/hook_context/pre_push.rb
```diff
@@ -18,17 +18,16 @@ def pushed_refs
...
-      @modified_files ||= Overcommit::GitRepo.modified_files(refs: ref_range)
+      @modified_files ||= pushed_refs.each_with_object(Set.new) do |pushed_ref, set|
+        set.merge(pushed_ref.modified_files)
+      end.to_a
```
taufek:
```
require 'benchmark'
require 'set'

iterations = 10
data = 100.times.map do
  Array.new(1_000) { rand(0..1_000) }
end

Benchmark.bmbm do |x|
  x.report('array.uniq') do
    iterations.times do
      data.uniq
    end
  end

  x.report('array.each_with_object{ merge }.to_a') do
    iterations.times do
      data.each_with_object(Set.new) do |datum, set|
        set.merge(datum)
      end.to_a
    end
  end
end

puts 'Are they equivalent?'
puts data.uniq == data.each_with_object(Set.new) { |datum, set| set.merge(datum) }.to_a

Rehearsal ------------------------------------------------------------------------
array.uniq                             0.250000   0.000000   0.250000 (  0.251087)
array.each_with_object{ merge }.to_a   0.140000   0.000000   0.140000 (  0.138263)
--------------------------------------------------------------- total: 0.390000sec

                                           user     system      total        real
array.uniq                             0.240000   0.000000   0.240000 (  0.244006)
array.each_with_object{ merge }.to_a   0.130000   0.000000   0.130000 (  0.138962)
Are they equivalent?
false
```

taufek: Seems like using `Set` with `each_with_object` is faster than `Array` with `uniq`.

lib/overcommit/hook_context/pre_push.rb
```diff
@@ -18,17 +18,14 @@ def pushed_refs
     end
 
     def modified_files
-      @modified_files ||= Overcommit::GitRepo.modified_files(refs: ref_range)
+      @modified_files ||= pushed_refs.map(&:modified_files).flatten.uniq
```
taufek:
```
require 'benchmark'
require 'set'

iterations = 10
data = 100.times.map do
  Array.new(1_000) { rand(0..1_000) }
end

Benchmark.bmbm do |x|
  x.report('array.flatten.uniq') do
    iterations.times do
      data.flatten.uniq
    end
  end

  x.report('array.each_with_object{ merge }.to_a') do
    iterations.times do
      data.each_with_object(Set.new) do |datum, set|
        set.merge(datum)
      end.to_a
    end
  end
end

puts 'Are they equivalent?'
puts data.flatten.uniq.sort == data.each_with_object(Set.new) { |datum, set| set.merge(datum) }.sort

Rehearsal ------------------------------------------------------------------------
array.flatten.uniq                     0.100000   0.010000   0.110000 (  0.097410)
array.each_with_object{ merge }.to_a   0.140000   0.000000   0.140000 (  0.148073)
--------------------------------------------------------------- total: 0.250000sec

                                           user     system      total        real
array.flatten.uniq                     0.090000   0.000000   0.090000 (  0.100320)
array.each_with_object{ merge }.to_a   0.140000   0.000000   0.140000 (  0.151008)
Are they equivalent?
true
```

Seems like for this case, calling `flatten.uniq` is fast.

spec/overcommit/hook_context/pre_push_spec.rb
```diff
@@ -121,7 +169,7 @@
           `git add . 2>&1 > #{File::NULL}`
           `git commit -m "Update Branch 2" 2>&1 > #{File::NULL}`
 
-          should include(*%w[added-1 update-me added-2].map { |file| File.expand_path(file) })
+          should == %w[added-1 added-2 update-me].map { |file| File.expand_path(file) }
```
sds: One thing worth noting is that using `==` here is slightly more brittle, as it prevents us from changing our implementation (which could return the file names in a different order) without breaking tests.

At the end of the day, our goal is to make sure we're returning the correct _set_ of files, not files in a specified order, which is why we used `include`. However, perhaps you changed this because we weren't checking for unintended files to be included. In that case, you could use a combination of `include` along with verifying the count to make sure no additional files are in the set.

This is not a big deal, so I'm fine merging this as-is, but it's important we're not overly-specific in tests.

sds:
Thanks @taufek!

spec/overcommit/hook_context/pre_push_spec.rb
taufek: I’ll take a note for future reference. Thanks.

PR #553, opened 2018-02-12
Extract Out PreCommit Message Methods to Utils

taufek:
Extracted messages methods in PreCommit context to Utils module.

This is a step closer to make these PreCommit hooks usable on other context.

### DIFF
--- a/lib/overcommit/hook/pre_commit/base.rb
+++ b/lib/overcommit/hook/pre_commit/base.rb
@@ -1,4 +1,5 @@
 require 'forwardable'
+require 'overcommit/utils/messages_utils'
 
 module Overcommit::Hook::PreCommit
   # Functionality common to all pre-commit hooks.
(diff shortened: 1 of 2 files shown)


lib/overcommit/utils/messages_utils.rb
```diff
@@ -0,0 +1,82 @@
...
+      # @raise [Overcommit::Exceptions::MessageProcessingError] line of output did
+      #   not match regex
+      # @return [Array<Message>]
+      def extract_messages(output_messages, regex, type_categorizer = nil)
+        output_messages.map.with_index do |message, index|
+          MessageProcessor.new(message, regex, type_categorizer, index).process
+        end
+      end
+
+      MessageProcessor = Struct.new(:message, :regex, :type_categorizer, :index) do
```
sds: It's not clear to me why we're creating an inner class via a `Struct` that shares the same name as the [`MessageProcessor`](https://github.com/brigade/overcommit/blob/77832e556bc2113badb6079d7cd3d6e7d6e9a3aa/lib/overcommit/message_processor.rb#L10) class, even if it is isolated. Rather than create a struct and call `process` on it, why not just get rid of the struct and run:

```ruby
# Replace `MessageProcessor.new(message, regex, type_categorizer, index).process` with:
process_message(message, regex, type_categorizer, index)
```

...and redefine the `process` method below to take arguments.

This allows you to have the remaining methods in this file as regular methods. They can remain `private` and isolated from the outside world, since you're using `class << self` above.

lib/overcommit/hook/pre_commit/base.rb
```diff
@@ -1,79 +1,11 @@
...
-    # @param regex [Regexp] regular expression defining `file`, `line` and
-    #   `type` capture groups used to extract file locations and error/warning
-    #   type from each line of output
-    # @param type_categorizer [Proc] function executed against the `type`
-    #   capture group to convert it to a `:warning` or `:error` symbol. Assumes
-    #   `:error` if `nil`.
-    # @raise [Overcommit::Exceptions::MessageProcessingError] line of output did
-    #   not match regex
-    # @return [Array<Message>]
-    def extract_messages(output_messages, regex, type_categorizer = nil)
```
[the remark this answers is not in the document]
taufek: Sounds good to me.

lib/overcommit/utils/messages_utils.rb
taufek: I do agree it is a poor choice for naming the Struct. 

The reason why I created an object to process each message within the loop is to avoid passing the context to the methods. 

```
extract_type(match, message)

extract_file(match, message)

extract_line(match, message)
```

If we move the methods to an object at individual message level we will methods without argument because now all the previous arguments are accessible via instance variables. 


```
extract_type

extract_file

extract_line
```

This way we distinguish the methods between the list level (messages) and indivual item level ( a message)

But if you still prefer to keep the methods with arguments I will change it.

lib/overcommit/hook/pre_commit/base.rb
```diff
@@ -1,79 +1,11 @@
 require 'forwardable'
+require 'overcommit/utils/messages_utils'
```
taufek: @sds ,

I have a question on this file loading. It is possible to avoid defining this and use autoloading instead? I notice modules like GitRepo and FileUtils we can use them via autoloading.

lib/overcommit/utils/messages_utils.rb
sds: I can see the reason why using a `Struct` can save you from having to pass arguments explicitly, but I don't think this really gets you much, as it [hides the dependencies](https://medium.com/@copyconstruct/small-functions-considered-harmful-91035d316c29#1aa7) of the method (i.e. what it actually needs to operate).

Method calls with arguments passed to them give you much more information about what the method is actually doing, rather than having it read a number of instance variables internal to the object.

lib/overcommit/hook/pre_commit/base.rb
sds: As we are not using autoloading anywhere else in this project, I would prefer not to introduce it at this time. I've been bitten by odd loading errors when trying to be too clever (usually not problems caused by autoloading itself, but by how others load a library locally). If someone uses your code in a way you didn't expect, they could trigger "autoloading" of classes at a time that you don't expect, which can cause difficult to debug issues.

lib/overcommit/utils/messages_utils.rb
taufek: Understood.

lib/overcommit/hook/pre_commit/base.rb
taufek: Noted.

taufek:
@sds ,

This is ready for re-review. Thanks for the feedback.

sds:
Thanks for addressing those comments, @taufek!

Issue #554, opened 2018-02-13
"Directory not empty" error after upgrading 0.41 → 0.42: 


mattbrictson:
After upgrading overcommit from 0.41.0 to 0.42.0, the first time I attempt to use `git commit` on an overcommit-enabled project I get this crash:

```
$ git commit
Directory not empty @ dir_s_rmdir - /Users/mbrictson/Code/myapp/.git/hooks/old-hooks
/Users/mbrictson/.rbenv/versions/2.5.0/lib/ruby/2.5.0/fileutils.rb:250:in `rmdir'
/Users/mbrictson/.rbenv/versions/2.5.0/lib/ruby/2.5.0/fileutils.rb:250:in `block in rmdir'
/Users/mbrictson/.rbenv/versions/2.5.0/lib/ruby/2.5.0/fileutils.rb:249:in `each'
/Users/mbrictson/.rbenv/versions/2.5.0/lib/ruby/2.5.0/fileutils.rb:249:in `rmdir'
/Users/mbrictson/Code/overcommit/lib/overcommit/installer.rb:142:in `preserve_old_hooks'
/Users/mbrictson/Code/overcommit/lib/overcommit/installer.rb:58:in `update'
/Users/mbrictson/Code/overcommit/lib/overcommit/installer.rb:20:in `run'
.git/hooks/pre-commit:67:in `<main>'
Report this bug at https://github.com/brigade/overcommit/issues
```

Digging in a bit further, I found that 3c37254 in 0.42.0 made a trivial change to the git hooks. Thus, the first time overcommit 0.42 is used on a existing overcommit-enabled project, it will attempt to upgrade to the new version of the git hooks.

It backs up any existing hooks to `old_hooks_path`, and then reaches this line:

```ruby
# Remove old-hooks directory if empty
FileUtils.rmdir(old_hooks_path)
```

For me, this fails with a "Directory not empty" error. Seems like there should be a guard clause here that skips the `rmdir` if the directory is not empty?

https://github.com/brigade/overcommit/blob/838a571852ac43e782dba58264f542c9a8a9d913/lib/overcommit/installer.rb#L129-L143

sds:
Hey @mattbrictson, thanks for discovering this. This is really weird, as our tests should have picked this up.

[...]

PR #555, opened 2018-02-13
Add Ruby 2.5.0 to build matrix

mattbrictson:
Bug #554 was specific to Ruby 2.5.0, which is a good reminder that 2.5.0 should be included in the Travis matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,14 @@ rvm:
   - 2.2
   - 2.3.6
   - 2.4.3
+  - 2.5.0
 
 before_script:
   - git config --global user.email "travis@travis.ci"
   - git config --global user.name "Travis CI"
 
 before_install:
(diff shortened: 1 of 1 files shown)


mattbrictson:
I based this PR on 838a571 (i.e. prior to the #554 fix) just to prove to myself that Travis would have caught the bug. I'll rebase this on master once Travis completes.

mattbrictson:
... and it succeed in Travis anyway. I guess it must be a macOS-specific issue.

In any case, it is still probably worthwhile to add 2.5.0 to the build matrix.

Rebased.

sds:
Thanks for the PR, @mattbrictson. Very strange that you couldn't reproduce on Travis, but having 2.5 in the build matrix is a good thing regardless, so thanks!

PR #570, opened 2018-04-03
git rev-parse gives wrong result in git 1.17.10

Antronin:
Strangly enough the --common-git-dir option just returns "--git-common-dir" with git version 1.17.10 if no $GIT_COMMON_DIR defined. The only way I can get correct answer is to stick to --git-dir.

### DIFF
--- a/lib/overcommit/utils.rb
+++ b/lib/overcommit/utils.rb
@@ -59,7 +59,9 @@ module Overcommit
       def git_dir
         @git_dir ||=
           begin
-            result = execute(%w[git rev-parse --git-dir])
+            cmd = %w[git rev-parse]
+            cmd << GIT_VERSION < '2.5' ? '--git-dir' : '--git-common-dir'
+            result = execute(cmd)
             unless result.success?
               raise Overcommit::Exceptions::InvalidGitRepo,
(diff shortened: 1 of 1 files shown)


lib/overcommit/utils.rb
```diff
@@ -59,7 +59,7 @@ def repo_root
       def git_dir
         @git_dir ||=
           begin
-            result = execute(%w[git rev-parse --git-common-dir])
+            result = execute(%w[git rev-parse --git-dir])
```
sds: This breaks support for [`git worktree`](https://git-scm.com/docs/git-worktree) (shame on me for not adding a proper test explicitly using worktrees in 10c20808).

In the meantime, as long we are running Git 2.5+ we should use `--git-common-dir`, otherwise we can use `--git-dir`. You can express this as:

```ruby
cmd = %w[git rev-parse]
cmd << GIT_VERSION < '2.5' ? '--git-dir' : '--git-common-dir'
result = execute(cmd)
```

sds:
Thanks @Antronin. Made some amendments and merged in 23b8942944db00b3590db90be1ecd3a9cd231f22.

PR #572, opened 2018-04-10
Added min_subject_width to TextWidth

namuit:
I added the support of `min_subject_width` in `TextWidth` with default value as `0` to obtain a behavior like the actual one.

```yaml
  TextWidth:
    min_subject_width: 0
```

A developer can set the `min_subject_width` to improve the commit messages in the team, for example a co-worker can avoid the `EmptyMessage` check putting as subject a single char.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -108,6 +108,7 @@ CommitMsg:
     enabled: true
     description: 'Check text width'
     max_subject_width: 60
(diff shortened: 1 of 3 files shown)


sds:
Thanks @namuit! Merged in fc03ddef with some amendments in 9b0956ba. Appreciate your contribution!

PR #575, opened 2018-04-20
Credo: only lint applicable files

prem-prakash:


prem-prakash:
Hi @sds some chance of a release so we can use it ?

Issue #579, opened 2018-05-17
Command not executed


damien-biasotto:
So first of all, very nice tool :)

This issue is regarding an upcoming PR, I'm trying to create a hook for PHP-Cs-Fixer : 

Before I submit that pull-request, I tried it on one of my project. Somehow the hook says  everything is fine, but the file was not touched.

If I run the same command manually, the file is updated. I'm not sure how to debug that (I'm brand new in the ruby world). and what I'm doing wrong  and I could use your help.

sds:
Hey @damien-biasotto, glad you're finding value with the tool.

Overcommit does not yet support pre-commit hooks with file-changing side effects. See #238 for details. Hope that explains what you're observing!

damien-biasotto:
Ah indeed that makes  sense :)

PR #580, opened 2018-05-17
PhpCsFixer Pre-Commit hook

damien-biasotto:
I tried to implement a php-cs-fixer pre-commit hook.

But somehow it seems that the command is not executed. I opened an issue there: https://github.com/brigade/overcommit/issues/579


sds:
Thanks!

PR #590, opened 2018-07-29
Stylelint linter added in pre-hook

Ruffeng:
This PR is in response with #514, which @brianbancroft started a PR but never had a chance to finish it.
The main problem with this linter was the kind of output that Stylelint was giving. Overcommit wasn't ready to handle multiline errors for the same file. So I asked if it could be possible to do a compact format for that [stylelint#3415](https://github.com/stylelint/stylelint/issues/3415). 

Thanks to @ntwb, that was possible ([PR 3488](https://github.com/stylelint/stylelint/pull/3488)). Therefore, I was able to implement the new linter at overcommit. 

To run properly the linter in the CLI: `stylelint "**/*.css" -f compact`. I would really appreciate if someone can verify I added correctly the way to run it in `default.yml`

I would like to say that this PR is made based on @briancroft 's PR. He deserves part of the contribution. As well @ntwb to make it possible from the Style lint side 

Feel free to code review and verify everything is ok 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -546,6 +546,7 @@ issue](https://github.com/brigade/overcommit/issues/238) for more details.
 * [SlimLint](lib/overcommit/hook/pre_commit/slim_lint.rb)
 * [Sqlint](lib/overcommit/hook/pre_commit/sqlint.rb)
 * [Standard](lib/overcommit/hook/pre_commit/standard.rb)
+* [StyleLint](lib/overcommit/hook/pre_commit/style_lint.rb)
 * [TrailingWhitespace](lib/overcommit/hook/pre_commit/trailing_whitespace.rb)
 * [TravisLint](lib/overcommit/hook/pre_commit/travis_lint.rb)
 * [TsLint](lib/overcommit/hook/pre_commit/ts_lint.rb)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -546,6 +546,7 @@ issue](https://github.com/brigade/overcommit/issues/238) for more details.
 * [SlimLint](lib/overcommit/hook/pre_commit/slim_lint.rb)
 * [Sqlint](lib/overcommit/hook/pre_commit/sqlint.rb)
 * [Standard](lib/overcommit/hook/pre_commit/standard.rb)
+* [StyleLint](lib/overcommit/hook/pre_commit/style_lint.rb)
```
sds: Nit: If the actual project name is capitalized "Stylelint", then we should use that same capitalization, and have the file be named `stylelint.rb` appropriately.

config/default.yml
```diff
@@ -711,6 +711,17 @@ PreCommit:
     install_command: 'npm install -g standard'
     include: '**/*.js'
 
+  StyleLint:
+    enabled: true
+    description: 'Check styles with Stylelint'
+    require_executable: 'stylelint'
+    flags: ['-f compact']
```
sds: These need to be split into separate arguments `['-f', 'compact']`, unless the `stylelint` CLI accepts `-f compact` as a single argument, but that would be surprising.

config/default.yml
```diff
@@ -711,6 +711,17 @@ PreCommit:
     install_command: 'npm install -g standard'
     include: '**/*.js'
 
+  StyleLint:
+    enabled: true
```
sds: We do not enable hooks by default unless they are truly applicable to almost any project. Set this to `false`.

Ruffeng:
@sds Thanks for your comments. I hope now will be ok 👍

PR #605, opened 2018-10-28
Use Hash#fetch with block

HarlemSquirrel:
According to `fasterer`, "Hash#fetch with second argument is slower than Hash#fetch with block"

This provides a very small improvement but since this gem is used with various workflows I feel that any speed improvement is a win.

Before

```
Finished in 1 minute 12.6 seconds (files took 0.49573 seconds to load)
1343 examples, 0 failures
```

After

```
Finished in 1 minute 12.47 seconds (files took 0.48929 seconds to load)
1343 examples, 0 failures
```

### DIFF
--- a/lib/overcommit/configuration.rb
+++ b/lib/overcommit/configuration.rb
@@ -41,7 +41,7 @@ module Overcommit
       @concurrency ||=
         begin
           cores = Overcommit::Utils.processor_count
-          content = @hash.fetch('concurrency', '%<processors>d')
+          content = @hash.fetch('concurrency') { '%<processors>d' }
           if content.is_a?(String)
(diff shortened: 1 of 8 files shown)


sds:
The difference in timing of a tenth of a second for tests that take over a minute to run are likely due to random jitter rather than actual performance benefit. 

While I could see lazy evaluation bringing benefits with respect to fewer objects allocated, this seems like a sacrifice in readability for questionable benefit.

With all that said—I will merge if you get the CI build passing. Thanks!

HarlemSquirrel:
I have mixed feelings about this as well. The readability suffers most with blocks containing an empty hash.

PR #611, opened 2019-01-07
CI: Include Ruby 2.6.0 in build matrix

olleolleolle:
This PR extends the matrix.

It also tries to keep the Bundler version before 2.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ rvm:
   - 2.3.7
   - 2.4.4
   - 2.5.1
+  - 2.6
 
 before_script:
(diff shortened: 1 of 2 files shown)


olleolleolle:
Oh, it seems that only the AppVeyor build process is enabled.

olleolleolle:
This one also needs some undoing:

```
$ gem update --system
Updating rubygems-update
ERROR:  Error installing rubygems-update:
	rubygems-update requires Ruby version >= 2.3.0.
```

sds:
Thanks @olleolleolle! The AppVeyor builds are notorious flakey. This looks good to me.

PR #612, opened 2019-01-08
Add --force-exclusion flag to default Reek config

tjwallace:
Similar to the flag provided by Rubocop, the `--force-exclusion` flag ensures files excluded via the `exclude_paths` config option are still excluded when they are explicitly passed to the `reek` command.


sds:
Test failure was unrelated. Thanks!

PR #616, opened 2019-02-07
Terraform support

### LINKED ISSUE #567: A Terraform formatter on pre-commit would be useful
Frequently we see terraform .tf/tfvars files with diverging formats. Running `terraform fmt` on pre-commit would be awesome.

BTW, if I submit a plugin for this, would it be considered?

msurovcak:
Simply check `terraform fmt` to ensure Terraform code is ok.

Fixes #567

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -738,11 +738,11 @@ PreCommit:
     include: '**/*.swift'
 
   TsLint:
-      enabled: false
-      description: 'Analyze with TSLint'
-      required_executable: 'tslint'
-      install_command: 'npm install -g tslint typescript'
-      include: '**/*.ts'
+    enabled: false
+    description: 'Analyze with TSLint'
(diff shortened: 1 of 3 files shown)


config/default.yml
```diff
@@ -737,12 +737,19 @@ PreCommit:
     install_command: 'brew install swiftlint'
     include: '**/*.swift'
 
+  Terraform:
```
sds: I would say we should call this `TerraformFormat`, as that is ultimately what it is checking. While I know Terraform is a pretty ubiquitous tool at this point, I think for some users just seeing "Terraform" doesn't necessarily hint at what it is doing.

msurovcak:
Thanks for the review @sds, addresses both comments as they make perfect sense + updated error message to be more human language, not just filename.

PR #618, opened 2019-02-19
add cookstyle support to fix #617

### LINKED ISSUE #617: Support for CookStyle
It would be really nice to have support for Chef cookstyle.  I could submit a pull-request for it pretty easily, it's basically just a dialect of rubcop and I was able to get a test to work by just copying the RuboCop related stuff to CookStyle and it worked great.

indygwyn:


config/default.yml
```diff
@@ -230,7 +230,17 @@ PreCommit:
     install_command: 'npm install -g coffeelint'
     include: '**/*.coffee'
 
-  Credo:
+   CookStyle:
```
sds: Indentation looks slightly off here, which may explain why the tests failed.

indygwyn: isn't it ironic my updating of a style linter gets caught by a style linting problem

indygwyn:
Not sure how I can fix that check in AppVeyor that's failing 
`Check for broken symlinks............................[BrokenSymlinks] FAILED` ...um

sds:
I would ignore the symlink error in AppVeyor, @indygwyn.

I missed this in my last review, but can you add an accompanying test? You can see an example from another recent pull request here: https://github.com/brigade/overcommit/pull/616/files#diff-2da0444d3bba168fda49f24b7abe274d

indygwyn:
@sds Okay, I tried writing my own rspec's but I kept getting an error, so I reverted to just a copy of rubocop's which should work fine since cookstyle is just a wrapper around rubocop, hopefully this works for ya.

sds:
Thanks @indygwyn!

Issue #348, opened 2016-02-19
Pre-commit hook proposal: filesize


lencioni:
It would be nice to have a hook that warned about large files being committed. 1MiB might be a good default, but it could probably even be set to something smaller.

lencioni:
It would be nice to have a way to configure the hook to warn at one level and error at another level. For instance, if you wanted to warn at 512KiB but error at 1MiB, it would be nice if this hook could support that.

jawshooah:
What units should we use for config values? Should we be able to parse `512KiB`, `512 KiB`, `1MB`, `1MiB`, etc? Or should it just be in bytes?

lencioni:
I think we should stick with something simple, so I'd say that the unit should not be in the value (we can mention it in the config name for clarity). I'd expect most people to set these to something in the KiB->MiB range, so perhaps the value should be in KiB? What do you think?

sds:
My perspective is probably skewed because I deal so much with configuring systems where memory/file size constraints need to be specified explicitly, but I _really_ prefer having units in any sort of measure of size. It's never intuitive what the default units are, and different tools use different defaults.

I would strongly prefer we support units (perhaps even enforcing that units are included). Whether we want to support the difference between `KB` and `KiB` I'm more flexible on (but personally: long live the [IEC](https://en.wikipedia.org/wiki/International_Electrotechnical_Commission)!)

I could see this being reused elsewhere, so it would make sense to go in the `Overcommit::Utils` module.

lencioni:
@sds I think units are important but I was advocating them being used in the configuration name instead of the value. I don't really feel strongly either way.

davidstosik:
If anyone get here and is interested in a solution, this is what we're gonna try to use where I work:
```rb
module Overcommit::Hook::PreCommit
  class FileSize < Base
    DEFAULT_SIZE_LIMIT_BYTES = 1_000_000 # 1MB

    def run
      return :pass if oversized_files.empty?

      oversized_files.map do |file|
        error_message_for(file)
      end
    end

    private

      def oversized_files
        @_oversized_files ||= build_oversized_file_list
      end

      def build_oversized_file_list
        applicable_files.select do |file|
          File.size(file) > size_limit
        end
      end

      def size_limit_bytes
        config.fetch("size_limit_bytes", DEFAULT_SIZE_LIMIT_BYTES)
      end

      def error_message_for(file)
        Overcommit::Hook::Message.new(
          :error,
          file,
          nil,
[...]

PR #636, opened 2019-04-15
Implement FileSize pre-commit hook

davidstosik:
Coming from https://github.com/brigade/overcommit/issues/348

After a few modifications from what I posted in the issue, the result is now this:

image

Note the file limit is now dynamically output in the hook's description so that it does not need to be repeated in each FileSize error.

Feel free to check the commit history to have an idea of my thinking process.

Improvements such as supporting file size units are welcome, but please let me suggest to open another PRs once this one is merged. 🙇

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## master (unreleased)
 
 * Drop support for Ruby 2.2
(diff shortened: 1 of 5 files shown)


sds:
Thanks!

PR #645, opened 2019-06-09
Add more replace_branch config options

jethrodaniel:
Currently, [ReplaceBranch][1] only runs if the commit type is [commit][2],
which means the hook won't run if you are using a template, for
instance.

This commit enhances the replace_branch prepare commit message hook to
allow specifying the types of commits to not apply this hook to.

So now, I can run this hook during merges, and when using templates by excluding `template` and `merge` in the config

```
ReplaceBranch:
  enabled: true
  skipped_commit_types: ['message', 'squash']
```

[1]: https://github.com/sds/overcommit/blob/569dc24a64e4b839b247658a688deccb03b716b8/lib/overcommit/hook/prepare_commit_msg/replace_branch.rb#L7
[2]: https://github.com/sds/overcommit/blob/569dc24a64e4b839b247658a688deccb03b716b8/lib/overcommit/hook/prepare_commit_msg/replace_branch.rb#L10

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1192,6 +1192,7 @@ PrepareCommitMsg:
     description: 'Prepends the commit message with text based on the branch name'
     branch_pattern: '\A.*\w+[-_](\d+).*\z'
     replacement_text: '[#\1]'
+    skipped_commit_types: ['message', 'template', 'merge', 'squash']
     on_fail: warn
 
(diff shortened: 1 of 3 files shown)


sds:
Tagging @dcluna as a heads up, as they submitted the original `ReplaceBranch` implementation.

PR #648, opened 2019-06-21
[FEATURE] New pre-commit hook: RubySyntax

diegojromerolopez:
Pre-commit hook RubySyntax check ruby files for syntax errors and show where the errors are.

It uses ruby -c to check syntax.

Useful if some files are not loaded automatically when performing the tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -678,7 +678,11 @@ PreCommit:
     enabled: false
     description: 'Check ruby syntax'
     required_executable: 'ruby'
-    command: ['bash', '-c', 'find . -name "*.rb" -exec ruby -c {} \; | grep -q "Syntax OK"']
+    command: [
+      'ruby',
+      '-e',
+      'ARGV.each { |applicable_file| ruby_c_output = `ruby -c #{applicable_file}`; puts ruby_c_output if ruby_c_output != "Syntax OK" }'
+    ]
     include:
       - '**/*.gemspec'
       - '**/*.rb'
--- a/config/default.yml
+++ b/config/default.yml
@@ -678,6 +678,7 @@ PreCommit:
     enabled: false
     description: 'Check ruby syntax'
     required_executable: 'ruby'
+    command: ['bash', '-c', 'find . -name "*.rb" -exec ruby -c {} \; | grep -q "Syntax OK"']
     include:
       - '**/*.gemspec'
       - '**/*.rb'
--- a/config/default.yml
+++ b/config/default.yml
@@ -674,6 +674,14 @@ PreCommit:
       - '**/*.gemspec'
       - '**/*.rb'
 
+  RubySyntax:
+    enabled: false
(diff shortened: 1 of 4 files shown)


lib/overcommit/hook/pre_commit/ruby_syntax.rb
```diff
@@ -0,0 +1,28 @@
...
+module Overcommit::Hook::PreCommit
+  # Runs `ruby -c` against all Ruby files.
+  #
+  class RubySyntax < Base
+    MESSAGE_TYPE_CATEGORIZER = lambda do |type|
+      type.match?(/^(syntax)?\s*error/) ? :error : :warning
+    end
+
+    def run
+      cmd = ['bash', '-c', 'find . -name "*.rb" -exec ruby -c {} \; | grep -q "Syntax OK"']
```
sds: You should be using the `command` option in the `RubySyntax` configuration section of `config/default.yml` to set this command. This allows users to overwrite if they so choose (e.g. to use a different `find` / `grep` utility which may run faster for their purposes.

sds: Note that the way you've structured this means that the command will run against every `*.rb` file in your repository. `pre-commit` hooks should run against only the files that were modified. 

You should be using the `include` option of the `RubySyntax` hook configuration to specify this (which you do), so it's not clear to me why you are specifying the command you have here.

diegojromerolopez: OK, you're right. Solved in 1117218.

diegojromerolopez: Yes, you are right, my last commit fixes that. Only changed files will be checked.

diegojromerolopez:
The reason to use this pre-commit hook and not rubocop is having some legacy project with so bad-quality code that we assert that code syntax is right.

I suppose it could be achieved by tunning Rubocop options, but this way should be easier.

config/default.yml
```diff
@@ -674,6 +674,19 @@ PreCommit:
...
       - '**/*.rb'
 
+  RubySyntax:
+    enabled: false
+    description: 'Check ruby syntax'
+    required_executable: 'ruby'
+    command: [
+      'ruby',
+      '-e',
+      'ARGV.each { |applicable_file| ruby_c_output = `ruby -c #{applicable_file}`; puts ruby_c_output if ruby_c_output != "Syntax OK" }'
```
sds: Judging success based on outputted text is potentially brittle if Ruby ever changes the output for a failed parse.

More robust would be to depend on the exit status of the binary, as a status of `0` means "success" by convention. 

For example:

```ruby
ARGV.each { |applicable_file| output = `ruby -c #{applicable_file}`; puts output unless $?.success? }
```

sds:
Thanks!

PR #649, opened 2019-06-24
Relax childprocess dependency version

delphaber:
`childprocess` turned `1` on January, so I tried to relax `childprocess` dependency version from `~> 0.6` to `>= 0.6.3`.

Let's see if CI passes :)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,3 +13,5 @@ gem 'coveralls', '~> 0.8'
 
 # Pin RuboCop for Travis builds.
 gem 'rubocop', '0.54.0'
+
+gem 'ffi' if Gem.win_platform?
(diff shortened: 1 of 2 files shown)


delphaber:
Dang! It breaks on Windows! Maybe because of https://github.com/enkessler/childprocess/pull/132 ? 🤔

sds:
Thanks for the PR, @delphaber. 

Yes, `childprocess` requires the `ffi` gem to be installed on Windows machines. However, it doesn't require this by default since most use cases don't need it and it slows down installs. See https://github.com/enkessler/childprocess/issues/150 for context. 

I think the simplest workaround is to add the following to this project's `Gemfile` so we get the tests to pass on Windows:

```ruby
gem 'ffi' if Gem.win_platform?
```

What do you think?

delphaber:
Makes totally sense! thanks :) I'm updating the PR

delphaber:
Done! Let's wait for the CIs :)

sds:
Thanks for the PR! Fixed the broken test and merged in #653.

PR #657, opened 2019-07-08
Bug/652 markdown lint hook fails with unexpected output

### LINKED ISSUE #652: Markdown lint hook fails with unexpected output
With the following `.overcommit.yml`:
```
PreCommit:
  Mdl:
    enabled: true
```

And a `README.md`:
```
## Oops
```

Then attempting to commit will fail with:
```
Analyze markdown files with mdl.................................[Mdl] FAILED
Unexpected output: unable to determine line number or type of error/warning for output:

A detailed description of the rules is available at https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md
```

Running `mdl README.md` manually will show:
```
README2.md:1: MD002 First header should be a top level header

A detailed description of the rules is available at https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md
```

I guess this extra line in the output may be new in `mdl` and has broken the hook.

steinybot:
Fixes #652

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Add `skipped_commit_types` option to `ReplaceBranch` prepare-commit-msg hook
 * Add `RubySyntax` pre-commit hook
 * Relax `childprocess` dependency to allow version 1.x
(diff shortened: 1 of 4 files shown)


sds:
Thanks for the fix!

PR #660, opened 2019-07-30
Fix typo: `require_executable` -> `required_executable`

jfly:
It looks like this incorrect field name has existed since the inception
of the Stylelint plugin in
https://github.com/sds/overcommit/commit/22229c95854cdd02d19e6e24e20d7df9ac23fea3.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -759,7 +759,7 @@ PreCommit:
   Stylelint:
     enabled: false
     description: 'Check styles with Stylelint'
-    require_executable: 'stylelint'
+    required_executable: 'stylelint'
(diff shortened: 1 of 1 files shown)


sds:
Thanks!

ChrisBr:
@sds this triggers to update the signature. Maybe this should be part of the CHANGELOG because that's something people should know about, no?

sds:
Hey @ChrisBr, thanks for catching. This was an oversight.

Updated change log in 86ef69096245f298e7fb55ef7017fa6bd68511dc.

ChrisBr:
Thanks @sds 👍

PR #665, opened 2019-09-01
300 remember all previously verified configurations

steinybot:
Initial attempt at solving #300.

This works by creating git blobs for each part of the signature and then taking each object id (SHA1) and pairing that with the hash (SHA256) of the contents of that blob. Those pairs are then written to another blob, delimited by a space, and one pair per line.

When checking if a signature has been verified then we compare the contents of the signature blob and not just the hash.

We can't just rely on the object id (SHA1) as collision attacks are feasible. A malicious repo owner could use such an attack to create an object with the same SHA1 as the configuration and hook contents.

I have kept the `overcommit.configuration.signature` in the git config for now but I don't think it is necessary. Removing it would loose the ability to show the `No previously recorded signature for configuration file` error but that doesn't seem like a huge loss. It may still be useful for change detection which is what the previous `signature_changed?` method did but nothing actually used it for that purpose, it was for signature verification instead.

I haven't tried using this locally yet so it could do with some manual testing before merging.

There are quite a few tests that fail for me but they also fail on master. Will see what the tests do on this PR.

Let me know what you think.


lib/overcommit/signature.rb
```diff
@@ -0,0 +1,102 @@
...
+      # @return [String]
+      def sign(arg)
+        Digest::SHA256.hexdigest(arg)
+      end
+
+      # Calculates the signature of some _Hash_.
+      #
+      # @param arg [Hash]
+      # @return [String]
+      def sign_signatures(arg)
```
sds: Ditto on comment about arg name.

lib/overcommit/signature.rb
```diff
@@ -0,0 +1,102 @@
...
+        sign(to_blob(arg))
+      end
+
+      # Returns the object signature for some blob.
+      #
```
sds: Missing doc for the argument type.

lib/overcommit/signature.rb
```diff
@@ -0,0 +1,102 @@
...
+
+module Overcommit
+  # Calculates, stores, and retrieves stored signatures of git objects.
+  module Signature
+    class << self
+      # Calculates the signature of some _String_.
+      #
+      # @param arg [String]
+      # @return [String]
+      def sign(arg)
```
sds: Could there be a more descriptive name than the generic term `arg`?

sds:
Thanks for your reply. Apologies for the delay in response, as I am traveling. 

If I simplify what you're proposing down to to its core elements, it appears you are asking:

1. Whether to use public key-based signatures to utilize a trust model similar to PGP
2. Whether to use a different ref space to store signatures (such as the one used by `git-notes`)

These still seem more complicated than a simple file of SHA256 signatures, for benefits that don't seem to have strong justification. 

With respect to 1: a web of trust model adds complexity and is still [considered broken](https://medium.com/@bblfish/what-are-the-failings-of-pgp-web-of-trust-958e1f62e5b7).

With respect to 2: I've never used `git-notes`, but this seems like we are tying our implementation to another Git feature, rather than simply taking a hash of a configuration and storing that in a file. Furthermore, if it's in a ref space, could an attacker inject notes? Can you trust notes?

My personal experience running this project is that additional complexity needs to have strong reasoning in order to be included. Revoking a signed configuration should be simple (delete a line). Revoking all signed configurations should be simple (delete a file). 

If I've mischaracterized these suggestions, please correct me. At this point in time, I am still strongly in favor of someone taking #505 to completion.

PR #668, opened 2019-09-10
Add C to rubocop warnings

henghonglee:
C is for convention. it should be a warning instead of an error. 

This PR fixes it


henghonglee:
seems like it was expected

PR #682, opened 2019-10-13
Add Test on Hook Base "run?" Method

taufek:
Context
=
Add test before updating the method in other PR (https://github.com/sds/overcommit/pull/681).


sds:
Thanks for adding these tests!

PR #681, opened 2019-10-13
Introduce `exclude_branch_patterns` on Hook Config

taufek:
Context
=
This will allow user to skip any Hook based on branch patterns.

For example, if the user wish to skip a particular Hook (i.e. Rubocop) for `staging` branch,
the user could specify below

```
PreCommit:
  Rubocop:
   exclude_branch_patterns: ['staging']
```
or using glob partial pattern

```
PreCommit:
  Rubocop:
   exclude_branch_patterns: ['stag*']
```

By default `exclude_branch_patterns` is empty (includes all branches).

PR #683, opened 2019-10-13
Introduce "exclude_remote_names" to Skip Hook During PrePush Based on Remote Name

taufek:
Context
=
Allow user to skip certain Hooks when pushing changes to certain remote repository.

For example, user could a Hook (i.e. RSpec), when pushing to 'heroku' remote name.
```
PrePush:
  RSpec:
    exclude_remote_names: ['heroku']
```

This is an attempt to solve issue reported in https://github.com/sds/overcommit/issues/588

### DIFF
--- a/lib/overcommit/hook/base.rb
+++ b/lib/overcommit/hook/base.rb
@@ -286,6 +286,10 @@ module Overcommit::Hook
       @config['exclude_branch_patterns'] || []
     end
 
+    def exclude_remote_names
+      @config['exclude_remote_names'] || []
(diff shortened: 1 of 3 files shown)


sds:
Thanks for the PR, @taufek.

Can we get the tests passing?

Furthermore, can we change the behavior to allow specifying a pattern (similar to #681) so that it's more flexible? Thanks!

taufek:
@sds , this is ready for re-review. Thank you for the feedback.

PR #690, opened 2019-10-14
Introduce 'ignore_branch_deletions' Config for PrePush Hooks

taufek:
Context
=
There are scenario we would like to skip pre-push Hooks when deleting remote branch.

Below is an example of git command to delete remote branch.

```
git push origin :to-be-deleted-branch
```

When we set `ignore_branch_deletions` as true for a PrePush Hook,
it will be skipped during remote branch deletion.


taufek:
This is an attempt to solve below issue:
https://github.com/sds/overcommit/issues/501

sds:
Thanks @taufek! 

Can you rebase so the conflicts are resolved?

taufek:
@sds , I've rebased and ready for re-review.

PR #692, opened 2019-10-16
[PrePush::ProtectedBranches] Implement setting destructive_only per branch

ChrisBr:
`PrePush::ProtectedBranches` supports a `destructive_only` setting. However, this setting is globally valid for all protected branches. It makes sense to set this setting on a per branch basis. This PR implements the following syntax for the configuration of `ProtectedBranches`.

```yaml
PrePush:
  ProtectedBranches:
    enabled: true
    branches:
      - protected
      - protected_for_destructive_only:
         destructive_only: true
```

In case it is not set on a branch it will fall back to the global setting. Additionally it will use the setting of the first matching branch. 

Example:
```yaml
PrePush:
  ProtectedBranches:
    enabled: true
    branch_patterns:
      - releases/*
      - releases/v1:
         destructive_only: true
```

In this case, a push to `releases/v1` would be forbidden because `releases/*` would match which uses the global setting for `destructive_only`. The solution in this case would be to flip the order of the branch definitions.

Credit for this idea goes to @davidstosik 🎉

### DIFF
--- a/lib/overcommit/hook/pre_push/protected_branches.rb
+++ b/lib/overcommit/hook/pre_push/protected_branches.rb
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 module Overcommit::Hook::PrePush
-  # Prevents destructive updates to specified branches.
+  # Prevents updates to specified branches.
+  # Accepts a 'destructive_only' option globally or per branch
+  # to only prevent destructive updates.
   class ProtectedBranches < Base
     def run
       return :pass unless illegal_pushes.any?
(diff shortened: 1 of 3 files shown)


sds:
Thanks @ChrisBr. Can you also update [the docs for ProtectedBranches](https://github.com/sds/overcommit/blob/e7b890706cc489aaa33f9f583fff85c4004426a0/lib/overcommit/hook/pre_push/protected_branches.rb#L4) to mention this functionality?

Otherwise, looks reasonable to me. Once you transition this from a draft happy to merge. Thanks!

ChrisBr:
@sds updated the documentation 👍 

Thanks

ChrisBr:
That was quick, thank you very much @sds 🎉

ChrisBr:
Is there a plan to release a new version anytime soon?

sds:
[0.52.0](https://github.com/sds/overcommit/releases/tag/v0.52.0) has been released.

PR #706, opened 2020-02-25
Add pronto to pre_push hooks

povilasjurcys:
Currently it's possible to add pronto only as PreCommit hook. This PR allows to add pronto as PrePush hook too.

### DIFF
--- a/lib/overcommit/hook/pre_commit/base.rb
+++ b/lib/overcommit/hook/pre_commit/base.rb
@@ -10,7 +10,7 @@ module Overcommit::Hook::PreCommit
 
     def_delegators :@context, :modified_lines_in_file, :amendment?, :initial_commit?
 
-    protected
+    private
 
     def extract_messages(*args)
       Overcommit::Utils::MessagesUtils.extract_messages(*args)
--- a/lib/overcommit/hook/pre_commit/base.rb
+++ b/lib/overcommit/hook/pre_commit/base.rb
@@ -10,7 +10,7 @@ module Overcommit::Hook::PreCommit
 
     def_delegators :@context, :modified_lines_in_file, :amendment?, :initial_commit?
 
-    private
+    protected
 
     def extract_messages(*args)
       Overcommit::Utils::MessagesUtils.extract_messages(*args)
--- a/lib/overcommit/hook/pre_push/base.rb
+++ b/lib/overcommit/hook/pre_push/base.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'forwardable'
+require 'overcommit/utils/messages_utils'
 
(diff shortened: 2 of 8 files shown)


lib/overcommit/hook/pre_commit/base.rb
```diff
@@ -10,7 +10,7 @@ class Base < Overcommit::Hook::Base
 
     def_delegators :@context, :modified_lines_in_file, :amendment?, :initial_commit?
 
-    private
+    protected
```
sds: Can you explain why we're making this change when the previous `private` worked?

lib/overcommit/hook/pre_push/base.rb
```diff
@@ -15,6 +16,12 @@ def run?
         (include_remote_ref_deletions? || !@context.remote_ref_deletion?)
     end
 
+    protected
```
sds: Why are we using `protected`? There isn't a situation where another instance is attempting to call this method, so `private` is the correct level of access control. If I'm misunderstanding your intent, let me know.

lib/overcommit/hook/pre_commit/base.rb
povilasjurcys: `extract_messages` is used in child classes.

> private" means that only methods of this class can access the member. "protected" means that methods of this class, or of *derived classes*, can access the member.
source: https://answers.unity.com/questions/610093/what-is-the-difference-between-private-and-protect.html

Just wanted to handle this in more "correct" way. Ruby does dot handle `private` and `protected` very well anyways, but I personally use `protected` when I want to indicate that method might be overwritten in child class. Sadly, in practice you can do the same with `private` methods too.

Let me know if you prefer private more.

lib/overcommit/hook/pre_push/base.rb
povilasjurcys: same answer as for https://github.com/sds/overcommit/pull/706/files#r386066591

lib/overcommit/hook/pre_commit/base.rb
sds: You are referencing documentation for a language that is not Ruby. `protected`, in a nutshell, allows a method to be called by another instance of the same class or subclass. It is not necessary in this context.

From https://ruby-doc.org/core-2.5.3/Module.html#method-i-protected
>If a method has protected visibility, it is callable only where `self` of the context is the same as the method. (method definition or instance_eval). This behavior is different from Java's protected method. **Usually `private` should be used.**

Emphasis mine.

povilasjurcys: thanks for pointing this out. You're right. `private` it is

sds:
Thanks!

PR #710, opened 2020-03-20
Live output option for rspec command

horiaradu:
* Add new option live_output
* If passed, show output in real time, instead of waiting for the command to finish

fixes https://github.com/sds/overcommit/issues/583

### DIFF
--- a/lib/overcommit/subprocess.rb
+++ b/lib/overcommit/subprocess.rb
@@ -31,7 +31,7 @@ module Overcommit
 
         process = ChildProcess.build(*args)
 
-        out, err = assign_output_streams(process)
+        out, err = assign_output_streams(process, options)
 
         process.duplex = true if options[:input] # Make stdin available if needed
         process.start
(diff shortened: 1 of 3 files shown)


lib/overcommit/subprocess.rb
```diff
@@ -85,14 +82,28 @@ def win32_prepare_args(args)
...
         %w[out err].map do |stream_name|
           ::Tempfile.new(stream_name).tap do |stream|
             stream.sync = true
             process.io.send("std#{stream_name}=", stream)
           end
         end
       end
+
+      def build_result(process, out, err, options = {})
+        return Result.new(process.exit_code, '', '') if options[:live_output]
```
sds, line 100: This can potentially break any processing done by the hook. If the `exit_code` is non-zero, hooks can sometimes read the stderr/stdout to determine the kind of failure that occurred. Given all the different ways Overcommit hooks _could_ be used, we can't introduce a change that fundamentally breaks the behavior.

PR #715, opened 2020-04-15
skip_if

### LINKED ISSUE #714: `optional_executable` option
We share our overcommit configuration between multiple repos, but some repos need to enable certain hooks while others don't. We've gotten around this by writing custom hooks that pass if the executable doesn't exist (for our purposes it's `eslint` - some repos have it, some don't). What would be great is to have an `optional_executable` option which just passes if the file doesn't exist.

kddeisz:
Closes #714. Provide a similar option to `required_executable`, except that when the executable isn't present in the path then it automatically exits instead of failing.


sds:
Hey @kddeisz, thanks for opening the PR.

Understand the the use case you're aiming for, but this feel antithetical to the original purpose of the tool, which is to enforce consistency for those that use it.

With that said, one can see the potential value of this if organizations/teams opt in to it. However, if we're going to introduce this into Overcommit as base feature, it needs to be more flexible than what is provided in this PR. Allowing organizations/teams to decide _when_ a hook may be skipped is an interesting concept and warrants the full flexibility of options.

Would be open to a pull request that adds an option which executes a command which checks if the hook can be skipped. For example:

```yaml
MyHook:
  skip_if: ['bash', '-c', '! which my-executable']
```

The semantics should be that the hook is skipped if the command executed returns successfully.

What are your thoughts?

kddeisz:
@sds yeah that makes total sense. I've updated the PR to be a `skip_if` option instead.

kddeisz:
@sds not sure why those tests are failing but I think this PR is good to go

PR #719, opened 2020-06-12
Fix `YamlLint` hook.

ajgon:
Previously, when yamllint was failing, the exit code of the tool was
reflecting that. However, recently yamllint introduced `--strict`
parameter to do that. Without, it always returns success no matter if
lints were successful or failed. Which caused false-positives in
ovecommit, which was thinking that everything is okay.

This fix adds `--strict` to `yamllint` invocation, to bring back proper
behavior, and force hook to fail, when the tool fails.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Overcommit Changelog
 
+## master
+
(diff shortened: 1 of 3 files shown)


sds:
Thanks!

PR #725, opened 2020-08-08
Update ReplaceBranch hook to avoid running on `--amend` by default

jethrodaniel:
- changes default to not run on `--amend`
- changes the default pattern
- improves the documentation for this hook
- refactors this hook's specs to reduce mocking and increase coverage

### DIFF
--- a/lib/overcommit/hook/prepare_commit_msg/replace_branch.rb
+++ b/lib/overcommit/hook/prepare_commit_msg/replace_branch.rb
@@ -2,8 +2,32 @@
 
 module Overcommit::Hook::PrepareCommitMsg
   # Prepends the commit message with a message based on the branch name.
+  #
+  # === What to prepend
+  #
   # It's possible to reference parts of the branch name through the captures in
   # the `branch_pattern` regex.
+  #
+  # For instance, if your current branch is `123-topic` then this config
+  #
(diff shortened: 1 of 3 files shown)


jethrodaniel:
Annoyingly, CI is complaining about a rubocop violation that doesn't exist. likely due to my force-pushes on this branch.

lib/overcommit/hook/prepare_commit_msg/replace_branch.rb
```diff
@@ -17,21 +43,21 @@ def run
...
-        "#{new_template} #{old_contents}"
+        "#{new_template}#{old_contents}"
       end
 
       :pass
     end
 
     def new_template
-      @new_template ||= Overcommit::GitRepo.current_branch.gsub(branch_pattern, replacement_text)
+      @new_template ||= Overcommit::GitRepo.current_branch.gsub(branch_pattern, replacement_text).strip
```
sds, line 53: I believe RuboCop is complaining about this line, @jethrodaniel. Should be easy to address.

sds:
Overall looks great. Once we get that one RuboCop line fixed we'll be good to merge. Thank you!

jethrodaniel:
Thanks - seems I was using rubocop with my global config, instead of this project's config.

sds:
Thank you!

PR #730, opened 2020-09-26
Delegate `modified_lines_in_file` to context

Quintasan:
@sds This went under the radar since `modified_files` is delegated via `Overcommit::Hook::Base`.

Now that I tried writing a custom hook I noticed `modified_lines_in_file` is not available inside `Overcommit::Hook::CommitMsg::Base`. I decided to expose only the method I need but I think it's worth thinking about if we want to delegate all methods that come from included modules by default


Quintasan:
@sds ding

sds:
Thanks for the ping.

PR #731, opened 2020-10-02
Add pre commit hook for ERB Lint

djfpaagman:
[ERB Lint][1] is a linter for Ruby's ERB files. This PR adds a pre commit hook to run this linter. 

I think I covered everything needed to add a new hook, let me know if it needs work!

[1]: https://github.com/Shopify/erb-lint


sds:
Thank you!

PR #738, opened 2021-01-13
Fail on yaml lint errors

Gui13:
Fixes https://github.com/sds/overcommit/issues/737

### DIFF
--- a/lib/overcommit/hook/pre_commit/yaml_lint.rb
+++ b/lib/overcommit/hook/pre_commit/yaml_lint.rb
@@ -11,7 +11,10 @@ module Overcommit::Hook::PreCommit
       if result.success?
         :pass
       else
-        return [:warn, result.stdout]
+        if result.stdout.include? "error"
+          return [:fail, result.stdout]
+        else
+          return [:warn, result.stdout]
       end
     end
   end
--- a/lib/overcommit/hook/pre_commit/yaml_lint.rb
+++ b/lib/overcommit/hook/pre_commit/yaml_lint.rb
@@ -9,13 +9,11 @@ module Overcommit::Hook::PreCommit
       result = execute(command, args: applicable_files)
 
       if result.success?
-        :pass
+        :pass 
+      elsif result.stdout.include?("error")
+        [:fail, result.stdout]
       else
-        if result.stdout.include? "error"
-          return [:fail, result.stdout]
-        else
-          return [:warn, result.stdout]
-        end
+        [:warn, result.stdout]
       end
     end
   end
--- a/lib/overcommit/hook/pre_commit/yaml_lint.rb
+++ b/lib/overcommit/hook/pre_commit/yaml_lint.rb
@@ -15,6 +15,7 @@ module Overcommit::Hook::PreCommit
           return [:fail, result.stdout]
         else
           return [:warn, result.stdout]
+        end
       end
     end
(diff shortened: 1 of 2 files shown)


Gui13:
I don't really know why the build fails, it seems that it fails on all the tests.

I'm not good enough in Ruby to analyze the issue though.

lib/overcommit/hook/pre_commit/yaml_lint.rb
```diff
@@ -11,7 +11,10 @@ def run
       if result.success?
         :pass
       else
-        return [:warn, result.stdout]
+        if result.stdout.include? "error"
+          return [:fail, result.stdout]
+        else
+          return [:warn, result.stdout]
```
sds, line 17: You are missing an `end` statement after this. This is a syntax error, which is why the specs fail.

sds (changes requested):
Thanks for this PR, @Gui13. Can you fix the syntax error?

lib/overcommit/hook/pre_commit/yaml_lint.rb
Gui13: Outch this doesn't look good indeed. I corrected.

spec/overcommit/hook/pre_commit/yaml_lint_spec.rb
```diff
@@ -34,9 +34,22 @@
       )
     end
 
-    it { should warn }
+    it { should fail }
```
sds, line 37: Note that `fail` is a keyword in Ruby so this test will always error (see the failing TravisCI run).

You should use the `fail_hook` keyword here and below.

sds (changes requested):

spec/overcommit/hook/pre_commit/yaml_lint_spec.rb
Gui13: Ok, it is done. I admit ruby is not my main language. I took inspiration from the existing code in this matter.

Gui13:
@sds I have rebased on your current master just in case.

The three failing tests are out of my scope, so I don't know what else to do.

In any case, this is what we use right now in production: it correctly handles error messages from yamllint

lib/overcommit/hook/pre_commit/yaml_lint.rb
```diff
@@ -11,7 +11,11 @@ def run
...
+        end
       end
```
sds, line 19: The test failure is because the way this `if`/`else` is written could be simplified.

You can resolve by writing as:
```ruby
if result.success?
  :pass 
elsif result.stdout.include?("error")
  [:fail, result.stdout]
else
  [:warn, result.stdout]
end
```

Notice this is "flatter" than what is currently written.

sds:
Note that there are still CI failures that should be pretty easy to fix. Can you have a look?
```
lib/overcommit/hook/pre_commit/yaml_lint.rb:12:14: C: Layout/TrailingWhitespace: Trailing whitespace detected.
lib/overcommit/hook/pre_commit/yaml_lint.rb:13:36: C: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
```

Gui13:
OK @sds I did it!

sds:
Thank you for your help!

PR #740, opened 2021-01-29
Add pub test and flutter test pre-push hooks

esensar:


esensar:
Not sure why AppVeyor build failed. From error message, doesn't seem to be caused by changes made here, but I may be wrong.

sds (approved):
I think the Appveyor failure was unrelated, given it passed on Ruby 2.4. Thank you!

PR #742, opened 2021-02-08
Revise index-tags script

bmckeough-showcaseidx:
- use `git rev-parse --git-dir` instead of $GIT_DIR to identify the git directory
- base what files are included in tag generation on the files managed
by git. Exclude files not managed by git

### DIFF
--- a/libexec/index-tags
+++ b/libexec/index-tags
@@ -6,10 +6,12 @@
 
 set -e
 
-trap "rm -f $GIT_DIR/tags.$$" EXIT
-err_file=$GIT_DIR/ctags.err
-if ctags --tag-relative -Rf$GIT_DIR/tags.$$ --exclude=.git "$@" 2>${err_file}; then
-  mv $GIT_DIR/tags.$$ $GIT_DIR/tags
+dir="`git rev-parse --git-dir`"
+
+trap "rm -f $dir/tags.$$" EXIT
+err_file=$dir/ctags.err
+if git ls-files | ctags --tag-relative -L - -f "$dir/tags.$$" 2>${err_file}; then
+  mv $dir/tags.$$ $dir/tags
   [ -e ${err_file} ] && rm -f ${err_file}
 else
   # Ignore STDERR unless `ctags` returned a non-zero exit code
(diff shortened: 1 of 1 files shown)


bmckeough-showcaseidx:
I was prompted to make this change because the index-tags script was not working for me. The $GIT_DIR environment variable was not populated, and so the script failed with "cat: /ctags.err: No such file or directory". Note that because the script is executed in the background and output is redirected to temporary files I had to pause the process in order to view the error output. That is, when the overcommit IndexTag hook is triggered it does not report a failure or error.

Git versions: 2.30.0
OS: macOS Catalina (10.15.7)
Shell: zsh 5.7.1 (x86_64-apple-darwin19.0)

sds:
Thanks for the pull request, @bmckeough-showcaseidx.

Can you elaborate on _why_ you made the second change? Developers may want to have `ctags` run without filtering to the files tracked by Git because there may be auto-generated files/etc. that they still want to be able to jump to definitions for. Filtering to files that are tracked by Git prevents this workflow.

Happy to merge without the filtering.

bmckeough-showcaseidx:
I reverted the `ctags` command to recursively tag all files.

My best recollection of why I preferred limiting tagging to files managed by git is that I have several repositories with large node_modules directories. I don't rely on tags for those files, and so preferred not generating tags for them.

sds (approved):
I'm supportive of you adding support for making ignoring files not tracked by Git an option for the hook, if you would like to see that added.

sds:
Thanks!

bmckeough-showcaseidx:
@sds Would you be receptive to a PR that parameterizes the index-tags script executed by IndexTags? If so, do you have any pointers on the appropriate way to make IndexTags accept a parameter that specifies the location of the index-tags script?

I'm imagining the ability to configure Overcommit to execute an indexing script of my chosing for each repository in which I use Overcommit. I imaging Overcommit would still supply the default libexec/index-tags script, but that this could be overridden by the user.

sds:
Very receptive!

Have a look at: https://github.com/sds/overcommit/blob/3174250ea04e54922d0c12f08ec907d29758d58b/lib/overcommit/hook/shared/index_tags.rb#L10

Notice `execute_in_background` takes an array of arguments. You could construct additional arguments to pass in based on the `config` for the hook, and then modify the script to use those arguments.

PR #744, opened 2021-02-21
Don't capture false positives

### LINKED ISSUE #743: eslint rule will fail due to filepaths that contain either "Warning" or "Error"
hi there,

my team uses overcommit to check our typescript codebase with eslint using the included [eslint rule](https://github.com/sds/overcommit/blob/master/lib/overcommit/hook/pre_commit/es_lint.rb). we discovered an issue with overcommit reporting an error on the eslint step, however running eslint by itself didnt show any errors. what we believe is happening is overcommit will parse the output of an eslint command that runs on a file or directory that contains the word "Error" or "Warning" as an actual error reported by eslint. the first case shows expected behavior, the second case shows the issue:

```bash
(base) apk@wolfenstein:~/Source/project(test-branch○) » touch error.ts
(base) apk@wolfenstein:~/Source/project(test-branch⚡) » git add error.ts
(base) apk@wolfenstein:~/Source/project(test-branch⚡) » git commit -m "test"
Running pre-commit hooks
Ensure eslint is used to lint JS.............................[EsLint] OK

✓ All pre-commit hooks passed

[test-branch b84f0904e] test
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 error.ts
```

```bash
(base) apk@wolfenstein:~/Source/project(test-branch○) » touch Error.ts
(base) apk@wolfenstein:~/Source/project(test-branch⚡) » git add Error.ts
(base) apk@wolfenstein:~/Source/project(test-branch⚡) » git commit -m "test"
Running pre-commit hooks
Ensure eslint is used to lint JS.............................[EsLint] FAILED
Unexpected output: unable to determine line number or type of error/warning for output:
$ yarn eslint --quiet --format=compact /Users/apk/Source/project/Error.ts
$ /Users/apk/Source/project/node_modules/.bin/eslint --quiet --format=compact /Users/apk/Source/project/Error.ts

✗ One or more pre-commit hooks failed
```

this line seems to be the culprit:

https://github.com/sds/overcommit/blob/bba5b75b2c16efccf2dbbbe4f2af103a15011119/lib/overcommit/hook/pre_commit/es_lint.rb#L24-L26

here is our .overcommit.yml for reference:

```yaml
EsLint:
    required_executable: "yarn"
    enabled: true
    command: ["yarn"]
    flags: ["eslint", "--quiet", "--format=compact"]
    description: "Ensure eslint is used to lint JS"
    exclude:
      - "/typings/**/*.d.ts"
    include:
      - "**/*.ts"
      - "**/*.tsx"
```
is it possible for overcommit to differentiate between what the current command running is, and the actual output of the command itself? that looks like the crux of the problem; it will parse both the command and it's output using the same rule.

adamvonbaron:
this is a pr to resolve #743. upon thinking about the cause of this problem more i realized that the output causing the false positive was output emitted by [yarn](https://yarnpkg.org/). my initial thought was to write a regex that would capture output that was yarn specific and ignore it, however this makes several assumptions about how other people may be using overcommit. i think it can be much more simple than that. the regex used to capture eslint messages here:
https://github.com/sds/overcommit/blob/6b25f5d60056614c58c5762a9c6e1fe8919742d2/lib/overcommit/hook/pre_commit/es_lint.rb#L33
actually does a really good job at capturing the data needed for overcommit to output information and adding a small tweak to that rule fixes the issue at hand. eslint messages have no whitespace before a file or filepath that has a linting issue, but something like
```bash
$ yarn eslint --quiet --format=compact /app/project/Error.ts
```
does. if we can just simply ignore any lines that DO have whitespace before the `<file>` capture group we can presume that line is not an eslint error or warning message, but some output emitted by some other program (such as yarn).

### DIFF
--- a/lib/overcommit/hook/pre_commit/es_lint.rb
+++ b/lib/overcommit/hook/pre_commit/es_lint.rb
@@ -19,20 +19,17 @@ module Overcommit::Hook::PreCommit
   # @see http://eslint.org/
   class EsLint < Base
     def run
+      eslint_regex = /^(?<file>[^\s](?:\w:)?[^:]+):[^\d]+(?<line>\d+).*?(?<type>Error|Warning)/
       result = execute(command, args: applicable_files)
       output = result.stdout.chomp
-      messages = output.split("\n").grep(/Warning|Error/)
+      messages = output.split("\n").grep(eslint_regex)
 
       return [:fail, result.stderr] if messages.empty? && !result.success?
       return :pass if result.success? && output.empty?
 
       # example message:
       #   path/to/file.js: line 1, col 0, Error - Error message (ruleName)
-      extract_messages(
-        messages,
-        /^(?<file>(?:\w:)?[^:]+):[^\d]+(?<line>\d+).*?(?<type>Error|Warning)/,
-        lambda { |type| type.downcase.to_sym }
-      )
+      extract_messages(messages, eslint_regex, lambda { |type| type.downcase.to_sym })
     end
   end
 end
--- a/spec/overcommit/hook/pre_commit/es_lint_spec.rb
+++ b/spec/overcommit/hook/pre_commit/es_lint_spec.rb
@@ -52,6 +52,18 @@ describe Overcommit::Hook::PreCommit::EsLint do
 
       it { should warn }
     end
+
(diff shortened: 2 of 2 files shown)


sds:
Thank you!

PR #745, opened 2021-03-16
Add ability for overcommit configs to inherit settings

tmandke:


sds:
Appreciate the pull request, @tmandke.

An inheritance feature would be useful, but a straight up `Hash#merge` call is perhaps too simple. We would likely want to implement some form of deep merge, or at least consider what other tools in the space (e.g. [RuboCop](https://github.com/rubocop/rubocop/blob/1e55b1aa5e4c5eaeccad5d61f08b7930ed6bc341/docs/modules/ROOT/pages/configuration.adoc#inheritance)).

Furthermore, some tests that assert the desired merge behavior should be included if we'd like to see this merged. Thank you!

PR #746, opened 2021-04-16
Disable post checkout for homebrew

quintrino:
Currently you can't install Homebrew taps when you've set overcommit as your default template as mentioned
here (https://github.com/sds/overcommit/issues/591) and here (https://github.com/Homebrew/brew/issues/4577) 

This PR detects when the current git repo is being used by Homebrew and disables overcommit.

### DIFF
--- a/spec/integration/disable_overcommit_spec.rb
+++ b/spec/integration/disable_overcommit_spec.rb
@@ -8,7 +8,10 @@ describe 'disabling Overcommit' do
   around do |example|
     repo do
       `overcommit --install > #{File::NULL}`
-      Overcommit::Utils.with_environment('OVERCOMMIT_DISABLE' => overcommit_disable) do
+      Overcommit::Utils.with_environment(
+        'OVERCOMMIT_DISABLE' => overcommit_disable, '
+        HOMEBREW_SYSTEM' => homebrew_system
+        ) do
         touch 'blah'
         `git add blah`
         example.run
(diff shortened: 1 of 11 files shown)


quintrino:
@sds I might need some help on figuring out how to test this, as I can't seem to wrap my head around how it's working with the existing tests and OVERCOMMIT_DISABLE.

sds (commented):
Thanks for the PR, @quintrino. Can I ask why we aren't making a change in the Homebrew project to unset the `GIT_TEMPLATE_DIR` environment variable? It seems like this would be a problem for anyone setting a template directory, not just Overcommit users.

Appreciate any clarity you can provide. Thank you.

quintrino:
As an update I'm trying to get this resolved at the Brew level before fixing it at the Overcommit level.

quintrino:
https://github.com/Homebrew/brew/issues/11364 is now merged into homebrew so this should resolve

https://github.com/Homebrew/brew/issues/4577
https://github.com/sds/overcommit/issues/591
(both of which were raised by @boardfish and helped me explain the issue to the Homebrew core team.

boardfish:
I'm afraid I've fallen a little out of touch with Overcommit since I originally raised that issue, but it's good to see this potentially getting fixed. Might need to jump back in with it soon. Thanks, @quintrino!

quintrino:
@broadfish Well I raised it in such a way that it addresses not just overcommit but all commit hooks so whether you're using overcommit or something else, either way you're covered. :)

PR #747, opened 2021-04-19
Add parsing of yamllint output

seanmil:
Parse the yamllint output so Overcommit has per-line awareness of
changes to support problem_on_unmodified_line option.


seanmil:
It isn't clear to me what impact this change had on the Appveyor test at `spec/integration/installing_overcommit_spec.rb`. They don't seem to be related.

sds:
Thank you!

PR #749, opened 2021-05-12
Support pre-commit.d folder

miry:
There are some tool that uses the pre-commit hook as well. It is possible to have multiple hooks in same time  with folder `.git/hooks/pre-commit.d`

### DIFF
--- a/template-dir/hooks/pre-commit
+++ b/template-dir/hooks/pre-commit
@@ -26,6 +26,14 @@ if hook_type == 'overcommit-hook'
   exit 64 # EX_USAGE
 end
 
+# Check if the hook located in .git/hooks/pre-commit.d
+parent_dir = File.basename(File.dirname($0))
+if parent_dir != "hooks"
+  if parent_dir.match?(/.d$/)
+    hook_type = parent_dir[..-3]
+  end
+end
+
 # Check if Overcommit should invoke a Bundler context for loading gems
 require 'yaml'
 # rubocop:disable Style/RescueModifier


sds:
Thanks for the PR, @miry. I'm not clear on the value of mixing different hook management systems in the same repository. 

Can you elaborate on why a development team would be using two systems?

miry:
@sds In some companies there are some internal tools exists for development envirnment. The internal tool required to have pre-commit hook, in same time it does not provide the same functionality as overcommit and focused only on specific commands.

As workaround I need manually to update the overcommit hook.

sds:
It's not clear why the specific nuances of some companies requires Overcommit to accommodate? It seems like a company should standardize on a single workflow? Imposing multiple workflows on the tools the company uses doesn't seem like the right course of action.

With that said—would be supportive of a pull request that allows users to **customize** the directories that Overcommit reads hooks from. That way you could specify _any_ directory (not just `pre-commit.d`).

miry:
The code that I introduce support any folder:

```
.git/hooks/<hook_type>.d
```

Before to update all <hook_type> templates, I wanted to be sure that it is valid idea. 

> It seems like a company should standardize on a single workflow? 

This is right, but it is not easy to do in big companies. In same time the tool support custom hooks as I had implemented. 
There is a comment about the approach to put custom hooks like overcommit and other hooks that overcommit should not cover in folder `.git/hooks/pre-commit.d` [comment](https://stackoverflow.com/questions/26624368/handle-multiple-pre-commit-hooks)

sds:
It's still not entirely clear how this adds support for _any_ directory. It adds support for directories of the form `<hook-type>.d`.  That is a limited set of possible directory names.

Could you not just register each hook as [described in the documentation](https://github.com/sds/overcommit#adding-existing-git-hooks)?

miry:
> Could you not just register each hook as described in the documentation?

I could not. I put overcommit to `.git/hooks/pre-commit.d` and updated the script to support folder name.

I think it is only my problem, so will keep fork version.

miry:
FYI: Added support of installation hooks to directories in the fork: https://github.com/miry/overcommit/pull/1

PR #753, opened 2021-05-28
Fixes #751 - Add `staged_changes?` check

### LINKED ISSUE #751: Deletes all files if we try to git commit without staging files
overcommit deletes all files if we try to commit without adding the files

Steps to reproduce:

```
1. Change a file
2. Do not `git add`
3. Try to commit 
    git commit -m "commit message"

4. Notice all the files have been deleted and staged 
```

The issue can be reproduced on this repository:

https://github.com/deepakmahakale/overcommit_issue

deepakmahakale:
Fixes #751

Add `staged_changes?` check to avoid stashing files if no changes have been staged.

stashing the files with no staged changes ends up deleting all files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Overcommit Changelog
 
+* Add `staged_changes?` check to avoid stashing files if no changes has been staged.
+
 ## 0.57.0
 
 * Fix `CommitMsg` hooks to be able to call `modified_lines_in_file`.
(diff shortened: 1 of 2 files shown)


sds:
As you can see from the tests, this won't work.

In order to guarantee you are running checks against files that are staged or otherwise are the view that Git sees, we always need to stash any unstaged changes in tracked files.

The solution to this particular problem appears to be removing the `--quiet` flag from the call to `git stash pop ...`. I'll do that in another PR.

sds:
See 648383ff3cd9d51ab83de87f2bf64fa9a0d6d559.

Issue #750, opened 2021-05-19
Teach ad-hoc hooks line-parsing using regexes in YAML


guillaume-d:
Many non-ad-hoc (pre-commit) hooks seem just to get the error code and parse stdout using `extract_messages` with the right `MESSAGE_REGEX` and `MESSAGE_TYPE_CATEGORIZER`.

Ad-hoc hooks could be taught how to do that from the overcommit YAML config file using:
- one regex for the message (a good default being the Emacs format)
- one optional regex to test whether the message is a warning or an error ("warning:" seems pretty common)
- maybe one optional regex to prefilter lines (empty lines, non-standard line, etc.), and flags to capture stdout and/or stderr, if we really need or want to handle more cases

Of course I guess in the end it all boils down to whether one prefers Ruby to regexes! :wink:
Anyway this might still be a good alternative for non-Ruby programmers to easily contribute hooks, and a good way to write less code (although I am not sure yet how tests could be written).

Thoughts?

sds:
Supportive of reducing the friction to integration a separate executable/script with Overcommits line handling functionality. Pull request with tests welcome!

guillaume-d:
OK, thanks for confirming!

Another question: is there a good reason why built-in hooks could not "be ad-hoc" like plugin hooks?
If there is no particular reason I'd like to change that too: it would make the new ad-hoc code easier to test: in the same PR one could change some built-in hooks to have an ad-hoc implementation, relying on the corresponding existing hook specs for (regression) testing.
(In that case probably all 3 `hook_loader/base.rb`, `built_in_hook_loader.rb`, and  `plugin_hook_loader.rb` would need changing.)
Also doing the above would add some test coverage for the ad-hoc code which does not look covered at all at the moment.

As regards more targeted unit testing I cannot find an existing unit test which exercise the class(es) I want to modify: only `cli_spec.rb` tests HookRunner, which itself uses the hook loader class(es) I would like to test.
Ideally one could reuse the existing hooks' spec code, somehow only override the overcommit configuration with ad-hoc hook definitions for a few hooks and synthetize new spec tests from that, but my actual rspec skills are much lower than my hand-waving ones here... :wink:

So would only indirect testing as outlined above be fine?

sds:
I could see many built-in hooks converted to the ad hoc format, yes. I haven't considered the implications too deeply, however, but would be open to a proposal.

[...]

PR #757, opened 2021-06-11
750: Teach ad-hoc hooks line parsing

guillaume-d:
Include some (partly regression) tests and a small doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,7 @@ Option                                  | Description
 `install_command`                       | Command the user can run to install the `required_executable` (or alternately the specified `required_libraries`). This is intended for documentation purposes, as Overcommit does not install software on your behalf since there are too many edge cases where such behavior would result in incorrectly configured installations (e.g. installing a Python package in the global package space instead of in a virtual environment).
 `skip_file_checkout`                    | Whether to skip this hook for file checkouts (e.g. `git checkout some-ref -- file`). Only applicable to `PostCheckout` hooks.
 `skip_if`                               | Array of arguments to be executed to determine whether or not the hook should run. For example, setting this to a value of `['bash', '-c', '! which my-executable']` would allow you to skip running this hook if `my-executable` was not in the bin path.
+`ad_hoc`                                | *["Ad-hoc" line-aware command hooks](#adding-existing-line-aware-commands) only.*
 
 In addition to the built-in configuration options, each hook can expose its
 own unique configuration options. The `AuthorEmail` hook, for example, allows
@@ -670,6 +671,67 @@ to understand which arguments are passed to the script depending on the type
 of hook, see the [git-hooks documentation][GHD].
 
 [GHD]: https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks
+
+### Adding Existing Line-Aware Commands
+
+Or in other words "low-code error format support."
+
+If you use tools that analyze files and report their findings line-by-line,
+and that Overcommit does not yet support, you may be able to integrate them
+with Overcommit without writing any Ruby code in a similar way as
+[for existing Git hooks](#adding-existing-git-hooks).
+
+These special line-aware command hooks behave and are configured the same way
+as the Git ones, except only file arguments get passed to them.
+Also they must have the `ad_hoc` option, so that, using the command output:
+- differentiating between warnings and errors becomes possible
+- modified lines can be detected and acted upon as defined by
+  the `problem_on_unmodified_line`, `requires_files`, `include` and `exclude`
+  [hook options](#hook-options)
+
+**Warning**: Only the command's standard output stream is considered for now,
+*not* its standard error stream.
+
+To differentiate between warning and error messages,
+the `warning_message_type_pattern` suboption may be specified:
+the `type` field of the `message_pattern` regexp below must then include
+the `warning_message_type_pattern` option's text.
+
+The `message_pattern` suboption specifies the format of the command's messages.
+It is a optional [(Ruby) regexp][RubyRE], which if present must at least define
+a `file` [named capture group][RubyRENCG].
+The only other allowed ones are `line` and `type`, which when specified
+enable detection of modified lines and warnings respectively.
+
+**Note**: The default value for this option is often adequate:
+it generalizes the quasi-standard [GNU/Emacs-style error format][GNUEerrf],
+adding the most frequently used warning syntax to it.
+
+For example:
+
+```yaml
+PreCommit:
+  CustomScript:
+    enabled: true
+    command: './bin/custom-script'
+    ad_hoc:
+      message_pattern: !ruby/regexp /^(?<file>[^:]+):(?<line>[0-9]+):(?<type>[^ ]+)/
+      warning_message_type_pattern: warning
+```
+
+**Tip**: To get the syntax of the regexps right, a Ruby interpreter like `irb`
+can help:
+
+```ruby
+require('yaml'); puts YAML.dump(/MY-REGEXP/)
+```
+
+Then copy the output line text as the YAML option's value, thereby
+omitting the `---` prefix.
+
+[RubyRE]: https://ruby-doc.org/core-2.4.1/Regexp.html
+[RubyRENCG]: https://ruby-doc.org/core-2.4.1/Regexp.html#class-Regexp-label-Capturing
+[GNUEerrf]: https://www.gnu.org/prep/standards/standards.html#Errors
 
 ## Security
 
--- a/lib/overcommit/hook_loader/base.rb
+++ b/lib/overcommit/hook_loader/base.rb
@@ -22,6 +22,34 @@ module Overcommit::HookLoader
 
     private
 
+    # GNU/Emacs-style error format:
+    AD_HOC_HOOK_DEFAULT_MESSAGE_PATTERN =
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -240,6 +240,7 @@ Option                                  | Description
 `install_command`                       | Command the user can run to install the `required_executable` (or alternately the specified `required_libraries`). This is intended for documentation purposes, as Overcommit does not install software on your behalf since there are too many edge cases where such behavior would result in incorrectly configured installations (e.g. installing a Python package in the global package space instead of in a virtual environment).
 `skip_file_checkout`                    | Whether to skip this hook for file checkouts (e.g. `git checkout some-ref -- file`). Only applicable to `PostCheckout` hooks.
 `skip_if`                               | Array of arguments to be executed to determine whether or not the hook should run. For example, setting this to a value of `['bash', '-c', '! which my-executable']` would allow you to skip running this hook if `my-executable` was not in the bin path.
+`ad_hoc`                                | *["Ad-hoc" line-aware command hooks](#adding-existing-line-aware-commands) only.*
```
sds, line 243: It seems like there should be a way to infer this without needing to be explicit. You'll also notice that we don't use the term "ad hoc" anywhere else in the project, except a few select comments. It's a bit of a lazy shortcut to referring to this concept.

It would perhaps be cleaner to simply offer high-level settings which change the behavior of the hook automatically. For example, if you specify `message_pattern`, perhaps that's indicator enough that it's an ad hoc hook.

Alternatively, perhaps we need only warn on when settings are specified which conflict. For example, specifying `message_pattern` when a hook file matching the name of the hook already exists.

Have you considered this approach? What are your thoughts?

README.md
```diff
@@ -671,6 +672,67 @@ of hook, see the [git-hooks documentation][GHD].
...
+
+```ruby
+require('yaml'); puts YAML.dump(/MY-REGEXP/)
+```
+
+Then copy the output line text as the YAML option's value, thereby
+omitting the `---` prefix.
+
+[RubyRE]: https://ruby-doc.org/core-2.4.1/Regexp.html
+[RubyRENCG]: https://ruby-doc.org/core-2.4.1/Regexp.html#class-Regexp-label-Capturing
```
sds, line 733: Let's link to at least Ruby 2.7 docs if possible (Ruby 2.4 is has reached [EOL](https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/))
https://ruby-doc.org/core-2.7.3/Regexp.html
https://ruby-doc.org/core-2.7.3/Regexp.html#class-Regexp-label-Capturing

lib/overcommit/hook_loader/base.rb
```diff
@@ -22,6 +22,34 @@ def load_hooks
 
     private
 
+    # GNU/Emacs-style error format:
+    AD_HOC_HOOK_DEFAULT_MESSAGE_PATTERN =
+      /^(?<file>(?:\w:)?[^:]+):(?<line>\d+):[^ ]* (?<type>[^ ]+)/.freeze
```
sds, line 27: Have we thought about not providing a default?

There are many potential hooks where all that matters is a pass/fail. If we can piggyback on the existing support for ad hoc hooks to simply run a specified script, and rather than try to offer a default we simply ensure that the user provides a `message_pattern` regex. This would likely reduce confusion since it forces the user to explicitly define the format rather than expect it to Just Work™.

README.md
```diff
@@ -671,6 +672,67 @@ of hook, see the [git-hooks documentation][GHD].
...
+adding the most frequently used warning syntax to it.
+
+For example:
+
+```yaml
+PreCommit:
+  CustomScript:
+    enabled: true
+    command: './bin/custom-script'
+    ad_hoc:
```
sds, line 717: Rather than having this high-level key, let's flatten this list so that if `command` is detected to be a script one can optionally specify `message_pattern` and other options to tweak how the output of that script is processed.

This would fit with the path forward discussed earlier, where built-in hooks change their implementations to use the `message_pattern` option rather than explicitly define their own message extraction ([example](https://github.com/sds/overcommit/blob/7d94fa55a74ceea705645d565a34ecefaca08ea5/lib/overcommit/hook/pre_commit/rubo_cop.rb#L20-L32) would simply be transformed into the invocation of a command and filter the output using the defined `message_pattern`, though perhaps RuboCop specifically would still need a custom one. Hopefully the idea is clear).

sds (commented):
Thanks so much for opening this pull request, @guillaume-d. Very excited to move this forward.

Left a few comments to clarify how we could potentially make this experience a seamless one without having to introduce the concept of an "ad hoc" hook to the configuration format. Curious to get your perspective.

README.md
guillaume-d: Sure, I just wanted to match [overcommit's current prerequisites](https://github.com/sds/overcommit/#requirements). So probably that section should at some point be updated too.

README.md
guillaume-d: IIUC to sum it up you'd like a flat YAML structure where the new `message_pattern` option's presence activates message extraction. If that's fine with you could we agree on the other details first? I would be glad if I could avoid too many rounds of PR rework.

README.md
guillaume-d: > It seems like there should be a way to infer this without needing to be explicit.

I think so too, I just wanted a clear separation from the existing stuff to get me started on writing the new code (I am still kind of a Ruby rookie, esp. WRT rspec).

> You'll also notice that we don't use the term "ad hoc" anywhere else in the project, except a few select comments. It's a bit of a lazy shortcut to referring to this concept.

Yeah, I noticed that! I wasn't sure why, and now I am not sure if you mean that the term "ad-hoc" should be avoided in general. Having a few precise terms help writing the doc a tad bit easier though (less repetitions/paraphrasing)...

> It would perhaps be cleaner to simply offer high-level settings which change the behavior of the hook automatically. For example, if you specify `message_pattern`, perhaps that's indicator enough that it's an ad hoc hook.

Agreed, but see further for another candidate option.

> Alternatively, perhaps we need only warn on when settings are specified which conflict. For example, specifying `message_pattern` when a hook file matching the name of the hook already exists.

Good idea, we could that in addition to the previous idea (flattening).
 
> Have you considered this approach? What are your thoughts?

I agree with everything you wrote, the fact that an empty option does mean something is reason enough to rethink the format!

lib/overcommit/hook_loader/base.rb
guillaume-d: I am not sure if by piggybacking you mean I should merge the code of both kinds of ad-hoc hooks? I'd rather keep it as it is now, else RuboCop will rightfully complain about too high metrics!

About not having a default for `message_pattern`: I empathize with your "too much magic" concern, although I am quite fond of this versatile default regex (not my baby, I think I stole it from some existing hook). Maybe we could keep it only as a good example/template in the docs.
If your only concern is having some option to detect message extraction, maybe we could use another one instead: maybe (as initially envisioned in #750) we could introduce `extract_messages_from` with a mandatory value (only `stdout` for now, but `stderr` would be trivial to add) and use that?

README.md
sds: You got it—a flatter YAML structure would be slightly easier to remember, and doesn't require the introduction of the word "ad hoc" to the configuration.

lib/overcommit/hook_loader/base.rb
sds: >Maybe we could keep it only as a good example/template in the docs.
Totally fine with that!

I hadn't considered the need to potentially decide between `stdout`/`stderr`. Slight preference in this case would be to create a wrapper script that sets up the redirects appropriately, rather than requiring Overcommit to address this issue. For example:

```yaml
# .overcommit.yml
PreCommit:
  CustomScript:
    enabled: true
    required_executable: './bin/custom-script.sh'
```

```bash
# custom-script.sh
#!/bin/bash

bundle exec rubocop 2>&1
```

Thoughts on leaving out `extract_messages_from` given the above is an option?

sds (commented):
Sincere apologies for the very slow response, @guillaume-d. Let me know your thoughts!

README.md
guillaume-d: So leaving the version as it is and resolving the conversation, unless you object of course.

guillaume-d:
> Sincere apologies for the very slow response, @guillaume-d. Let me know your thoughts!

No problem, was started to worry something bad may have happened.
In the meantime, I have been implementing some kind of flattening, I will just finish it (hopefully today) and push that for now as a stepping stone, and once done will have another look at implementing your suggestions.

README.md
guillaume-d: OK I now started using the term "line-aware" more systemically internally too for my new type of ad-hoc hooks, and "ad-hoc" is nowhere to be seen anymore in the configuration and docs.

lib/overcommit/hook_loader/base.rb
guillaume-d: I thought needing wrappers would be a non-Unix-portability issue for the project (only Windows without WSL or such nowadays I guess... and there's always PowerShell), but that's fine with me, I can live with wrappers and Windows being at a disadvantage! ;-)

I'll address that in the next iteration.

guillaume-d: As requested, `extract_messages_from` is now gone, and `message_pattern` no longer has a "magic" default value.

PR #764, opened 2021-10-12
Support hook type option to `--run`

elebow:
Sometimes, especially during hook development, you want to force running of a certain type of hook.

This adds an option to the `--run` parameter for specifying a hook type. The default behavior is unchanged.


sds (approved):
Tests pass locally on macOS. Thanks!

PR #766, opened 2021-10-29
Fix encoding of process output under windows

rymut:
Output of commands under windows is not UTF-8 by default, this
can lead to "invalid byte sequence in UTF-8" error on symlink check

### DIFF
--- a/lib/overcommit/subprocess.rb
+++ b/lib/overcommit/subprocess.rb
@@ -51,7 +51,7 @@ module Overcommit
         err.rewind
         out.rewind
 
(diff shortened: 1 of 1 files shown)


rymut:
Thanks @sds: in my case only calling win32_symlink is reproducible - but the underlying problem will still be present.
My current commit should fix the issue with braking tests under linux - if the problems with the tests will be still persisting then I will isolate a change to win32_simlink.

sds (approved):
Thank you for making the changes. Tests pass locally for me on macOS.

PR #767, opened 2021-11-12
Stylelint Pre-Commit: Fixed Message Regex

bert-mccutchen:
On the rare occasion messages from Stylelint do not match properly with the existing Regex. This change fixes the issue by matching up until the first colon.

Example:
`/app/assets/stylesheets/_example.scss: line 291, col 1, error - Unexpected duplicate selector ".example:first-child:after", first used at line 282 (no-duplicate-selectors)`

Produces:
Screen Shot 2021-11-12 at 11 44 46 AM


sds:
Thank you!

PR #777, opened 2022-04-25
Partial log flushing to make long hooks intuitive

rickmark:
Today a long hook will not print any output until it is complete.

This change introduces flushing (when available) to the status lines so that a hook will output before its run

### DIFF
--- a/lib/overcommit/logger.rb
+++ b/lib/overcommit/logger.rb
@@ -31,6 +31,11 @@ module Overcommit
       log
     end
 
+    # Flushes the [IO] object for partial lines
+    def flush
+      @out.flush if @out.respond_to? :flush
+    end
+
     # Write a line of output.
     #
     # A newline character will always be appended.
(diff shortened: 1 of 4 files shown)


sds:
RSpec 3 failure looks spurious.

rickmark:
> RSpec 3 failure looks spurious.

Which part?  My machine uses `main` for initial branch - which blew up the tests
I also got errors on `Bundler` thus adding the require
Is there a failure I didn't catch?

sds:
See the [failed Ruby 3.0 Windows build](https://github.com/sds/overcommit/runs/6215374390?check_suite_focus=true). No action required from you—was merely commenting that we are merging in spite of what appears to be a spurious failure.

rickmark:
Ah - yeah - I admit - I didn't bother with a windows test run - would you like a patch-set to do CI on PRs?  (We do this in https://github.com/hack-different/apple-knowledge/blob/main/.github/workflows/jekyll_ci.yml to run Jekyll but it could easily run `spec`)?

rickmark:
OM... just realized its a closed PR so I wasn't seeing the checks - carry on

PR #779, opened 2022-04-27
Fix pronto hooks swallowing generic errors

Drowze:
Found this issue when we were running overcommit on the context of Gemfile that didn't contain `pronto` on it. Ops!


sds:
Thanks for the contribution and for adding tests.

Agreed—failure looks spurious.

PR #796, opened 2022-09-08
Allow overriding Gemfile.lock

xjunior:
Allow overriding the Gemfile.lock file (respect BUNDLE_GEMFILE env var).


xjunior:
It looks like `Tests / rspec (2.6, windows) (pull_request)` sometimes times out, and the coverage went down by 0%.

that-jill (approved):
<3 thanks for fixing this and making our workflows easier

wadewinningham (approved):

PR #799, opened 2022-11-17
Add optional additional local configuration

julesros:
Adds support for an additional configuration file, `.local-overcommit.yml`, in order to allow local configuration on a local computer. That way the main config `.overcommit.yml` can be commited to a repo and a local developer can also add configurations to this new file that do not have to live in the main config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,10 @@ PreCommit:
     command: ['bundle', 'exec', 'rubocop'] # Invoke within Bundler context
 ```
 
+Additionally, you may wish to have repo-specific configurations that are local to your computer that are not part of the shared repo config.
+Adding a `.local-overcommit.yml` file in the top-level directory of the repository adds another configuration file. This file works the same as `.overcommit.yml`.
+Adding this to ignored files in a git repo will allow you to have a local configuration per repo.
+
 ### Hook Options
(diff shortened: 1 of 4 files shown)


julesros:
@sds (or other maintainers) Any quick insight into why all builds are failing? I see another recent PR with similar behavior in CI. Locally, the specs pass so I am not sure where to begin with debugging.

sds:
Sorry for the delay. The failures seem unrelated to your change (I get the same failures without your changes). 

With that said, I won't be able to dig into what causes them. Since this seems straightforward and passes locally, will go ahead and merge.

I will offer a word of caution: allowing users to "override" the behavior of overcommit with a configuration custom to them negates the benefit of having a shared configuration for a team. I would suggest that one think carefully before utilizing such a feature if one wants to have consistent hook results among a team.

Thanks for the PR!

julesros:
Thanks for the feedback and merge! Agreed regarding being cautious of overriding--our particular case was some engineers wanting to run the `IndexTags` hooks or other personal preference hooks with overcommit more as addition to the standard we'd like to enforce repo wide.

I've tried out the new release on my project and it seems to be working as expected.
Thanks again!

PR #805, opened 2023-01-10
Rescue Psych::DisallowedClass in YamlSyntax

HarlemSquirrel:
This improves the output for errors safe loading disallowed classes with psych 5.0

It would be nice to offer a way to specify `permitted_classes` somewhere. https://ruby-doc.org/3.2.0/exts/psych/Psych.html


sds:
Thanks for this. The failures seem unrelated to your change (I get the same failures without your changes). Will go ahead and merge.

PR #813, opened 2023-08-25
Add RSpec Hook to PreCommit, Add Applicable files if set

quorauk:
We recently added this functionality to one of our own Repos and thought we should share.

This change requires the 'include' parameter to be set before it uses the 'applicable_files' function. Is this the correct way to do this? Should we create a new parameter for this behaviour? 

Referenced Issue: https://github.com/sds/overcommit/issues/703


sds:
Thanks for the contribution! Sorry for the delay in merging. Fixed RuboCop errors and merged in #842.

PR #815, opened 2023-09-24
fix: local overcommit file merges with existing .overcommit.yml file

julesros:
The .local-overcommit.yml did not work as intended. When a repo level .overcommit.yml and .local-overcommit.yml both existed, only the latter applied. Further, the tests only checked that each file was loaded, not that the configs included the expected overrides.

This PR refactors:
- Specs:
  - When .overcommit.yml exists, applies settings and has default settings still
  - When both .local-overcommit.yml and .overcommit.yml exists has default settings and both files settings applied
- Logic to merge the .overcommit.yml file to default settings and if it exists, .local-overcommit.yml file to both

### DIFF
--- a/lib/overcommit/configuration_loader.rb
+++ b/lib/overcommit/configuration_loader.rb
@@ -70,11 +70,9 @@ module Overcommit
     # Loads a configuration, ensuring it extends the default configuration.
     def load_file(file, local_file = nil)
       overcommit_config = self.class.load_from_file(file, default: false, logger: @log)
-      if local_file
-        local_config = self.class.load_from_file(local_file, default: false, logger: @log)
-      end
+      l_config = self.class.load_from_file(local_file, default: false, logger: @log) if local_file
       config = self.class.default_configuration.merge(overcommit_config)
-      config = self.class.default_configuration.merge(local_config) if local_config
+      config = config.merge(l_config) if l_config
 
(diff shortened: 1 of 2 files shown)


julesros:
@sds Greetings! I was using the feature I added earlier this year and realized it had a bug. Providing a fix here for your consideration.

simoleone:
I just hit the same bug and found it rather confusing. Anything I can do to help get this merged?

julesros:
> I just hit the same bug and found it rather confusing. Anything I can do to help get this merged?
👋🏻 @simoleone , I need to take a look at where I left this. If it works as I had intended we can tag the maintainers for review I think. I'll try to get to it this weekend! If you'd like to test this branch locally as well and make sure it works as expected for the issue you found as well, that would probably be great.

simoleone:
hey! Yup this works the way I think it should 😄

simoleone:
Hey! Any way I can help out with this? 😄 . I know how free time goes with my own projects (or doesn't 🙃 ).

julesros:
> Hey! Any way I can help out with this? 😄 . I know how free time goes with my own projects (or doesn't 🙃 ).

Hi! Ah so sorry, yes, time is hard to find 🙃 
If you've pulled down this branch and it looks like it works, then I think we can just go ahead and ask for a review from maintainers and request it to be merged in. I had wanted to clean it up more but if it works we can go with it.

PR #818, opened 2024-01-10
Update RuboCop

andyw8:
The current version of Ruby is from April 2020. Since there are quite a lot of new issues identified, I've committed them to the `.rubocop_todo.yml` so that they can be addressed incrementally or disabled.


Gemfile
```diff
@@ -11,7 +11,7 @@ gem 'rspec', '~> 3.0'
 gem 'simplecov', '~> 0.21.0'
 gem 'simplecov-lcov', '~> 0.8.0'
 
-# Pin RuboCop for Travis builds.
-gem 'rubocop', '0.82.0'
+# Pin RuboCop for CI builds
```
andyw8, line 14: Changed since it looks like Travis is no longer used.

spec/overcommit/hook/pre_commit/phpcs_fixer_spec.rb
```diff
@@ -13,7 +13,7 @@
 
   context 'when phpcs fixer exits successfully with fixed file' do
     before do
-      # rubocop:disable Metrics/LineLength
+      # rubocop:disable Layout/LineLength
```
andyw8, line 16: RuboCop has changed the category of this cop.

PR #820, opened 2024-01-16
Allow childprocess 5

dentarg:
I think it only brings improvements: https://github.com/enkessler/childprocess/releases/tag/v5.0.0


sds:
Thanks!

PR #824, opened 2024-02-22
Get suite running smoothly again

benmelz:
Am planning on proposing/implementing a new feature, but first I noticed that there are a number of failing specs on main, so I figured I would start with that.

As written, this includes four main adjustments, two of which might actually be bugfixes:
- one spec was hanging indefinitely for me locally on a `Thread.join` call - explicitly capturing the child threads and joining them directly fixed it (might be machine dependent?)
- when an encoding is specified, the rails schema up to date hook was passing arguments incorrectly (kwarg containing `encoding` was being interpreted as the `length` positional arg) (possible bug)
- many specs were attempting to create file submodules without the `protocol.file.allow=always` config option causing failure
- alias detection for amendment appeared to be broken, due to regexp incompatibilities between ruby/git (git wasn't properly parsing `\s` in the value matcher for `git config --regexp [key] [value]` command), so moved that matching into ruby (possible bug, also maybe machine/git build opt dependent)
- dropped ruby 2.6 from the test matrix - it appears that it's far enough EOL that github has trouble even setting it up. I can attempt to add it back/get it working, but I figured it might be time to ditch it

*I don't have a windows machine, so I'm not gonna be much help with those CI builds (should note fails appear unrelated to anything I've done here).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,7 +14,6 @@ jobs:
       fail-fast: false
       matrix:
         ruby-version:
-          - '2.6'
           - '2.7'
           - '3.0'
           - '3.1'
--- a/lib/overcommit/hook/pre_commit/rails_schema_up_to_date.rb
+++ b/lib/overcommit/hook/pre_commit/rails_schema_up_to_date.rb
@@ -35,7 +35,7 @@ module Overcommit::Hook::PreCommit
     private
 
     def encoding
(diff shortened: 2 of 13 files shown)


sds (approved):
Thanks for this!

sds:
Would be open to dropping support for Ruby 2.7, given it's also EoL (but if we have tests passing, fine to keep).

PR #848, opened 2024-04-12
Support stylelint 16+, which emits messages on stderr

### LINKED ISSUE #823: Overcommit doesn't work with Stylelint 16
Stylelint 16 made a breaking change: lint failures are now printed on stderr instead of stdout.

Overcommit hasn't yet been updated to support Stylelint 16, and still expects the output on stdout. As a result, the pre-commit hook prints "OK" when there are actual lint failures.

Stylelint documentation regarding the breaking change: https://stylelint.io/migration-guide/to-16#breaking-changes

I suspect the best way to solve this is for Overcommit to check _both_ stderr and stdout, so that new and old versions of Stylelint are supported.

mattbrictson:
Before Stylelint 16, Stylelint emitted its messages on stdout. Hence Overcommit's Stylelint hook was built to read output from stdout.

Starting with Stylelint 16, Stylelint now emits its messages to stderr, which means the existing Overcommit implementation no longer works.

This PR updates Overcommit's Stylelint plugin to check for messages on _both_ stdout and stderr. That way <16 and >=16 versions of Stylelint are supported.

Fixes #823

### DIFF
--- a/lib/overcommit/hook/pre_commit/stylelint.rb
+++ b/lib/overcommit/hook/pre_commit/stylelint.rb
@@ -12,7 +12,7 @@ module Overcommit::Hook::PreCommit
 
     def run
       result = execute(command, args: applicable_files)
-      output = result.stdout.chomp
+      output = result.stdout + result.stderr.chomp
       return :pass if result.success? && output.empty?
 
       extract_messages(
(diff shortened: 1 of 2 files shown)


pilaf:
Hi @sds, are there any plans to release a new version with this change included soon?

Thanks!

sds:
Released in [0.64.0](https://github.com/sds/overcommit/releases/tag/v0.64.0)

pilaf:
@sds Awesome, thank you!

PR #853, opened 2024-08-10
 Expand tildes(~) for hooksPath

anakinj:
While installing overcommit git hooks i noticed that tildes are not expanded to home folders for the `hooksPath` config, the result was a tilde folder in the current directory.

This PR addresses that by switching `File.absolute_path` to `File.expand_path`. The [underlaying implementation](https://github.com/ruby/ruby/blob/v3_3_4/file.c#L3753) in Ruby is exactly the same just with this difference in how `~` is handled.


sds:
Thank you!

PR #854, opened 2024-08-10
Fix spec compatibility with recent git versions

anakinj:
The original array was defining the shellwords like this:
`["git", "commit", "-m", "\"Resolve", "conflicts\"", "-i", "some-file"]`

This combined with the most recent git version failed with:
```
error: pathspec 'conflicts"' did not match any file(s) known to git
```

This PR just simplifies the commit message to work with the %w array literal.


sds:
Thanks!

PR #859, opened 2025-01-23
Load bundled gems on expected version

xjunior:
Fixes https://github.com/sds/overcommit/issues/789

`psych` is no longer a default gem, but current and old ruby still ships with it. When we `require 'yaml'`, we activate that gem in whatever version that is bundled with ruby. Later on, we load bundler, and we `Bundler.setup`, which will then activate whatever version specified in the lock file. More often than not, they might not match.

The approach in this PR is to strip the `yaml` dependency completely from the hook scripts, while retaining the ability to configure the Gemfile using it.

### DIFF
--- a/template-dir/hooks/commit-msg
+++ b/template-dir/hooks/commit-msg
@@ -27,10 +27,11 @@ if hook_type == 'overcommit-hook'
 end
 
 # Check if Overcommit should invoke a Bundler context for loading gems
-config = File.read('.overcommit.yml')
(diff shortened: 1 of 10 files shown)


wadewinningham:
I have this issue and it resolved it for me.

Victor-Oliveira (commented):
Have this same issue too. Those changes resolved it for me.

sds:
Thanks!

sds:
Released in [0.65.0](https://github.com/sds/overcommit/releases/tag/v0.65.0).

PR #860, opened 2025-01-29
Add `-diff` cli option for running precommit hooks against diffs

benmelz:
For example, running `overcommit --diff main` from a feature branch will run pre-commit hooks against the diff between the two branches.

I was able to very easily leverage existing code for the bulk of the feature - this is mainly just adding the cli option, a hook context to do the execution and some tests based on the existing `--run-all` test.

---

For background, my team is responsible for a couple of really old, really large rails apps. Getting them completely in compliance with our various linters is a huge task that isn't getting done anytime soon (things are funky to the point that we've even observed breakages with "safe" auto-correct functions).

I introduced/started heavily encouraging overcommit so that we at least don't add _new_ linting offenses and things will naturally improve over time. It's been great, but offenses still slip through though here and there, especially with juniors who might be getting away with not having a local install (and/or abusing `OVERCOMMIT_DISABLE=1`).

An option like this would allow me to leverage the very useful "only apply to changed lines" logic within a ci environment and help enforce my desired "no new linting offenses" policy.

### DIFF
--- a/lib/overcommit/cli.rb
+++ b/lib/overcommit/cli.rb
@@ -9,6 +9,7 @@ module Overcommit
   class CLI # rubocop:disable Metrics/ClassLength
     def initialize(arguments, input, logger)
       @arguments = arguments
+      @cli_options = {}
       @input     = input
       @log       = logger
       @options   = {}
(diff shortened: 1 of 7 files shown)


sds (approved):
Nice. Thank you for the clean PR and tests.

PR #863, opened 2025-02-15
Restore ability to specify "gemfile: false" in config

### LINKED ISSUE #862: `gemfile:false` option no longer behaves as expected in YAML config
As of 0.65, when I run overcommit with `gemfile: false` in the config, I get the following error:

`Problem loading 'false': my-code-directory/false not found`

It seems like when #859 started using regex to parse the gemfile config instead of YAML, we ended up losing some nice YAML.parse functionality?

```ruby
# .overcommit.yml
gemfile: false

# hook code
config = File.read('.overcommit.yml') =~ /gemfile: ['"]?(.*)['"]?/
gemfile = Regexp.last_match(1) # gemfile is now set to the (truthy) string "false"

if gemfile # this is triggered with gemfile = 'false'
  ENV['BUNDLE_GEMFILE'] = gemfile

  begin
    Bundler.setup
  rescue Bundler::BundlerError => e
    puts "Problem loading '#{gemfile}': #{e.message}" # Problem loading 'false'
    puts "Try running:\nbundle install --gemfile=#{gemfile}" if e.is_a?(Bundler::GemNotFound)
    exit 78 # EX_CONFIG
  end
end
```

mattbrictson:
Setting `gemfile: false` in `.overcommit.yml` is supposed to disable Bundler. However, a recently-introduced bug causes `false` to be interpreted as the name of the gemfile. Bundler looks for a gemfile named "false", which fails, leading overcommit's hooks to crash.

This PR fixes the bug by adjusting the regex used to parse the `gemfile:` line in the config. Now, `false` is no longer interpreted as a gemfile name.

I added an integration test to verify the fix.

Fixes #862

### DIFF
--- a/spec/integration/gemfile_option_spec.rb
+++ b/spec/integration/gemfile_option_spec.rb
@@ -100,4 +100,59 @@ describe 'specifying `gemfile` option in Overcommit configuration' do
       end
     end
   end
+
+  context 'given a project that does not use a Gemfile' do
+    let(:hook) { normalize_indent(<<-RUBY) }
+      module Overcommit::Hook::PreCommit
+        class NoInvalidGemfileHook < Base
+          def run
+            if (gemfile = ENV["BUNDLE_GEMFILE"])
+              raise unless File.exist?(gemfile)
+            end
+
+            :pass
+          end
+        end
+      end
+    RUBY
+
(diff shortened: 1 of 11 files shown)


sds:
Thank you.

Issue #881, opened 2026-06-02
Possible race when parallel hooks read HookContext input


Felix-LeeSM:
Hi! I ran into a strange pre-push behavior while debugging a custom hook, and I think it comes down to a small race in `HookContext::Base#input_string`.

Right now it memoizes stdin like this:

```ruby
def input_string
  @input_string ||= @input.read
end
```

Since hooks are parallelized by default, two hooks that share the same context can call this at the same time. For hook types that get real data on stdin, like `pre-push`, that means both threads may try to read the same input stream.

The failure mode is subtle:

1. one thread reads the actual Git pre-push refs
2. another thread reads EOF and gets `""`
3. if the empty read is assigned last, `@input_string` stays cached as `""`

For `pre-push`, Git sends refs through stdin:

```text
<local ref> <local sha1> <remote ref> <remote sha1>
```

So if the empty string wins the race, helpers like `pushed_refs` can return an empty list even though Git did send refs. In my case this made a pre-push check silently skip the work it was supposed to do.

I think the fix can be pretty small: protect the first read with a mutex, so the stream is only consumed once and every hook gets the same cached input.

Something like:

```ruby
def input_string
  return @input_string if defined?(@input_string)

  @input_mutex.synchronize do
    @input_string = @input.read unless defined?(@input_string)
  end

  @input_string
end
```

I also put together a test that avoids relying on flaky timing: it uses a fake input object that blocks the first `read`, starts a second thread, and fails if the underlying input gets read more than once.

I have a branch here if that helps:

https://github.com/Felix-LeeSM/overcommit/tree/fix/thread-safe-hook-input

Focused spec, RuboCop, and `bundle exec overcommit --run` pass locally. The full spec suite has a few unrelated failures around git alias amendment detection on my machine.

Would you be open to a PR for this?

sds:
Thanks for the fix in #882!

PR #882, opened 2026-06-02
Make hook context input reads thread-safe

Felix-LeeSM:
## Summary

This fixes a small race in `HookContext::Base#input_string`.

Hook contexts currently cache stdin lazily with `@input_string ||= @input.read`. Since hooks are parallelized by default, two hooks sharing the same context can race during that first read. For hook types that receive real data on stdin, such as `pre-push`, one hook can consume the stream while another reads EOF and caches an empty string.

If that happens, helpers like `pushed_refs` can silently see no refs even though Git did provide them.

## Changes

- Add a mutex around the first `input_string` read so the stream is consumed once.
- Keep the cached string shared by all callers after the initial read.
- Add a spec that forces a concurrent read attempt without relying on flaky timing.

This follows up on #881.

## Verification

Focused checks pass locally:

```sh
bundle exec rspec spec/overcommit/hook_context/base_spec.rb
bundle exec rubocop lib/overcommit/hook_context/base.rb spec/overcommit/hook_context/base_spec.rb
bundle exec overcommit --run
```

I also ran the full spec suite. It still has a few unrelated failures around git alias amendment detection on my machine, but this change's focused spec passes.

### DIFF
--- a/lib/overcommit/hook_context/base.rb
+++ b/lib/overcommit/hook_context/base.rb
@@ -24,6 +24,7 @@ module Overcommit::HookContext
       @args = args
       @input = input
       @options = options
+      @input_mutex = Mutex.new
     end
 
     # Executes a command as if it were a regular git hook, passing all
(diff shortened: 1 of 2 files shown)


sds (approved):
Thanks!

PR #887, opened 2026-07-24
Fix operation state restoration in linked worktrees

xenyal:
## Context

I ran into this through a confusing shell prompt error after completing work in a linked worktree. Overcommit was restoring merge and cherry-pick state into the shared Git directory, which can make another checkout appear to be mid-operation. The shared directory is right for hooks, but these state files belong to the active worktree.

## Changes

- Resolve operation-state files through `git rev-parse --git-path`, with a fallback for Git versions before 2.5.
- Restore `MERGE_HEAD`, `MERGE_MODE`, `MERGE_MSG`, and `CHERRY_PICK_HEAD` in the active worktree.
- Preserve merge mode and message contents exactly.

### DIFF
--- a/lib/overcommit/git_repo.rb
+++ b/lib/overcommit/git_repo.rb
@@ -160,7 +160,10 @@ module Overcommit
         @merge_head = merge_head
       end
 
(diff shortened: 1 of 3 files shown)


sds:
Thank you

